PR #5, opened 2016-11-17
hypre dependency on MPI in configure

smithsg84:
hypre has a dependency on MPI so include MPI libs in configure test for working hypre library


smithsg84:
Looks OK and passing

PR #6, opened 2016-11-17
velocity pfb print support

jbeisman:


pfsimulator/parflow_lib/solver_richards.c
```diff
@@ -53,7 +53,7 @@
 typedef struct
 { 
    PFModule          *permeability_face;
-   PFModule          *phase_velocity_face;
+   // jjb PFModule          *phase_velocity_face;
```
smithsg84: When removing unneeded code, just delete it rather than commenting it out.   You probably followed some other parts of the code but we are getting rid of this.  It makes the code hard to follow after 4-5 different edits to things.  Thanks,

PR #45, opened 2017-10-25
Nc clm input

kulkarni1:
Hello Steve,
This branch contains following features
0) Time step 0 output containing static fields and initial pressure and saturation
1) All the parflow transient variables in netcdf output
2) All the clm variables in netcdf output. Separate file for netcdf
3) Meteorological forcing for CLM
The 2D variables are written as (time, y and x) and 3D variables are written as (time, z, y and x)
Little washita netcdf test case has been updated accordingly. User manual entries for necessary TCL flags and detailed usage is described.
Regards,
Ketan


pfsimulator/parflow_lib/parflow_netcdf.h
```diff
@@ -26,9 +25,13 @@
 #else
 #define MAX_NC_VARS 8192
 #endif
+#include<stdbool.h>
+
+static int ncID, xID, yID, zID, lev1ID, timID;
```
smithsg84: Why are static variables being declared in the header file?   This will create local copies in any file that #includes this header file.  Shouldn't these just all bin the .c file.

kulkarni1: Hi Steve,
You can reformat the code in my branch. I wanted global variables for dimension IDs in netcdf which would be accessible to all related netcdf routines and I do not have to pass too many arguments to functions. In theory they could be moved to individual .c files. I will check how to do it without breaking current working code.

kulkarni1:
Hi Steve,
I have moved the variable definitions from header to sources.

smithsg84:
Thanks,
Steve

PR #69, opened 2018-03-17
parFlowVR

xy124:
In the following I introduce changes to allow using the ParFlow model in a [FlowVR](flowvr.sourceforge.net) work flow facilitating to benefit from In Situ and In Transit analysis and file write. Especially the latter gives a remarquable performance improvement allowing file writes with only very little extra cost in wall time compared to a run without file write.

Furthermore we introduce steering. Thus the simulation can be changed online. This allows e.g. nudging, data assimilation and inverse modeling with minimal or even without file system accesses. Further this will allow interactive simulation runs in the future. In this version there is a VisIt connector already to inspect running simulations without file system interaction.

I would appreciate if you could tell me what you think of this and if you see the possibility to merge it in the main version of ParFlow. 

References/Inspirations:
- https://hal.inria.fr/hal-00941413/document
- https://hal.inria.fr/hal-00715252/document


xy124:
The configure build fails as I could not run all autotools to add `flowvr.c` and `flowvr.h` in `pfsimulator/parflow_lib`. Can someone please tell me the autotools version you are using? (My autoheader script fails always, even when running from within autoreconf in the master branch)

xy124:
@smithsg84 can you please tell me how you think about merging the features introduced in this branch?

smithsg84:
This branch is very out of date and can't be merged.   Closing issue.

Issue #81, opened 2018-06-07
Change in how Dirichlet bottom boundary conditions work (commit 1a066747)


cvoter:
I’ve been working with a version of parflow from around October 2016 (I don’t have good notes on the exact version) but recently updated to the current version on github. After doing so, I noticed a puzzling change in how my bottom boundary condition works that I’ve traced back to commit 1a066747 from May 1, 2017.

In the model where I observed a change, the domain has a constant DZ (0.1m) with ComputationalGrid.Lower.Z set to 0m. The bottom boundary conditions are set up as follows:

    pfset Patch.bottom.BCPressure.Type  DirEquilRefPatch
    pfset Patch.bottom.BCPressure.Cycle  constant
    pfset Patch.bottom.BCPressure.RefGeom  domain
    pfset Patch.bottom.BCPressure.RefPatch  bottom
    pfset Patch.bottom.BCPressure.alltime.Value  0.0

In versions of parflow before commit 1a066747, this results in a pressure head in the bottom layer that hovers around -0.05m, as I expect. In versions of parflow after commit 1a066747, the pressure head in the bottom layer is suddenly higher, closer to -0.025m. I *think* this is a bug in the newer versions, but I haven’t spent enough time studying the changed file (pfsimulator/parflow_lib/nl_function_eval.c) to spot exactly what the issue is. If it’s not a bug, then I’d appreciate help understanding the change that was made. Thanks in advance!

reedmaxwell:
Hi Carolyn-

Thanks for opening a tag, I also found this bug about a year ago and put in a request as well, though I don’t see it in the system.  If you swap your patch to the top, it will work, but of course this is just a workaround not a fix.

Reed

Reed M. Maxwell, Ph.D.
Rowlinson Professor of Hydrology
Department of Geology and Geological Engineering
Director, Integrated GroundWater Modeling Center
Colorado School of Mines
http://inside.mines.edu/~rmaxwell/

On Jun 7, 2018, at 2:35 PM, Carolyn Voter <notifications@github.com<mailto:notifications@github.com>> wrote:


I’ve been working with a version of parflow from around October 2016 (I don’t have good notes on the exact version) but recently updated to the current version on github. After doing so, I noticed a puzzling change in how my bottom boundary condition works that I’ve traced back to commit 1a06674<https://github.com/parflow/parflow/commit/1a066747d099984f2a0f7f1516f4908de8daad91> from May 1, 2017.

[...]

PR #83, opened 2018-06-20
Fix DirichletBC(z) (#81)

cvoter:
Please correct me if I'm wrong, but I think sep should not be divided by 2 for DirichletBC(z). All tests pass. Pressure heads in lowest layer now look as expected when apply a Dirichlet BC to bottom of domain.


reedmaxwell:
Thanks so much for the pull request and comment.  Laura and I reviewed your changes and think that at the bottom boundary DZ should be divided by two.  ParFlow places the boundary value on the edge of the cell and computes the flux to the center of the cell, so this is usually DX/2, DY/2, DZ/2.  However, can you provide more information about your test case and more details about how the pressures were looking incorrect for the old formulation.  Thanks!

cvoter:
Yep ok - I took another look at this a few months ago and I agree. I suspect my issue was that I matched the original error in my post-processing scripts so when it got fixed here, my post-processing scripts were then wrong.

PR #95, opened 2018-12-04
Issue 91 clm layers hardcoded

smithsg84:
Fixed duplicate keys for configuring CLM number of layers.


kulkarni1:
Bit late to comment on this issue. But this should not affect netCDF writing as I read the number of layers from SubGridNZ

PR #123, opened 2019-03-26
Added double quotation marks to CMakelists.txt to prevent erroneous i…

jbeisman:
…nsertion of semicolons during build


smithsg84:
Thanks for bugfix.

PR #144, opened 2019-06-11
Updating pftools.c

nbengdahl:
Adding a manual override option for the number of layers when calling pfdist

### DIFF
--- a/pftools/pftools.c
+++ b/pftools/pftools.c
@@ -124,6 +124,7 @@ int            PFDistCommand(
   int num_procs_y;
   int num_procs_z;
   int num_procs;
+  int nz_manual=0;
 
   Background    *background;
   Grid          *user_grid;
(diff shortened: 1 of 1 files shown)


pftools/pftools.c
```diff
@@ -133,9 +134,27 @@ int            PFDistCommand(
...
   char command[1024];
 
-  if (argc != 2)
+  // Setup and error checking for manual nz spec
+  if ((argc == 2)||(argc == 4))
   {
-    WrongNumArgsError(interp, LOADPFUSAGE);
+    if (argc == 4) /* Check that third argument is -nz */
+    {
+      if (strcmp(argv[2],"-nz")!=0)
```
smithsg84: Looking at TCL commands, I think it is more 'standard' to have the switches before the other args.

cmd ?switches? arg1 ... argn

Please move the -nz before the filename arg.

And in future the code change a documentation should be in a single pull request.

Thanks,
Steve

smithsg84: And can you modify one of the tests to use the flag....so we test it :)

Thanks
Steve

nbengdahl: Sure, I’ll get those changes made in my fork (just created, still learning!) then will probably submit a new pull request with all three changes (order of args, test case, and manual update)

Issue #163, opened 2019-07-25
Add HFB (Modflow) type condition between cells


LaurenLiloFoster:
A valuable add to ParFlow would be a simple interference condition between cells that the user specifies. This would be similar to the [HFB condition in Modflow](https://www.waterloohydrogeologic.com/help/vmod-flex/index.html?vm_define_wall_hfb.htm) or to the [itfc condition in FEHM](https://fehm.lanl.gov/docs/FEHM_UM_V3.3.0.pdf).

smithsg84:
Reed Maxwell comments
I did some looking this am (having spent a decent amount of time in PF solvers) and I think this is actually super quick to add.  If we read in three PFB files with multiplier values for X, Y and Z then all we need to do is multiply the fluxes in nl_function_eval and richards_jacobian_eval and the velocities.  For starters we could just set them all to 1 for the “False” setting and we would get the same behavior as normal, the user could set them something else.  The basic idea is that they multiply the fluxes at the cell interfaces, not the K values, so they work a little different than K since they aren’t averaged.  In NL function eval this could happen at lines 659-678 (see highlighted screen shot) or lines 729.  If at 729 they would look like this:
u_right = u_right*HFB_x[ip];
u_front = u_left*HFB_y[ip];
u_upper = u_upper*HFB_z[ip];

where HFB_x, y, z would be the multipliers with some convention that the value at ip would be really like an i+1/2,j,k — the same thing would be added into richards_jacobian lines 835, 854 and 869 where coeff would be multiplied by HFB_x, y and z respectively in each location (835 - x, 854 - y, 869 - z).

If the user just set these as constant by geom or read in by file HFB_x, y, z could be pretty straightforward to add (just three more problem_HFB_x type modules).

hope this helps / make sense

thanks!

Reed

khm293:
I second Lauren’s comment that this would be super useful. Thanks!

PR #167, opened 2019-08-08
Flow Barrier

reedmaxwell:
This creates a flow barrier capability in response to #163 this is equivalent to the hydraulic flow barrier (HFB) or flow and transport parameters at interfaces.  The flow barriers are placed at the fluxes as scalar multipliers between cells (at cell interfaces).  The following things were added to the code:

- three new vectors to store the FB values in X, Y and Z
- problem modules for each vector, (currently the values are only loaded as PFB files but this could be expanded in the future)
- modification to the function evaluation and jacobian for the flow barrier formulation
- two test cases in the test suite
- additions to the manual describing the physics and usage of this feature

The flow barrier boolean default to False, which sets the values to unity, thus providing backward compatibility.


parflow-docker:
The branch has conflicts and can't be merged. 

Since we already merged the this development branch once the normal workflow is to start a new branch that is current and so pull request from that.

PR #183, opened 2019-09-16
updating washita test

lecondon:
Updating the Washita PF-CLM test to reflect changes to drv_clmin.dat and to fix the start time.


lecondon:
Updated the Washita test run to reflect updates to the drv_clmin.dat. The starting hour changed to reflect the correct time zone for Washita and the correct outputs were  swapped out to reflect this.

PR #238, opened 2020-07-03
Fixed erroneous -D use in ParflowTest.cmake

ian-bertolacci:
ParflowTest.cmake as several occurrences of `-D={MPIEXEC_PREFLAGS=${MPIEXEC_PREFLAGS}` which causes MPIEXEC_PREFLAGS to never be used when running tests, which is an issue where MPIEXEC_PREFLAGS is necessary.

Fixed with `-DMPIEXEC_PREFLAGS=${MPIEXEC_PREFLAGS}`


ian-bertolacci:
Closed in favor of #239

PR #239, opened 2020-07-03
Fixes and upgrades to CMake testing system

ian-bertolacci:
I had to run and debug some build issues on a new system, and these changes allowed me to debug the issues and get it working on that system.

Fundamentally, the issue was that the proper mpiexec invocation (command *and* all specified flag arguments) when running some tests (particularly AMPS tests) was not being used, and so those tests failed.
This issue, and a number of others that I did not directly encountered but felt that the could be an issue in the future or for others, appears to be resolved.

I have tested this branch on several systems (including the one I originally had an issue with) and all tests for those configurations (all of which use the mpi1 amps layer, some version of OpenMPI, sequential IO, CLM, Hypre, and SILO) passed successfully.

These commits do the following
+ Fixed erroneous `-D={MPIEXEC_PREFLAGS=${MPIEXEC_PREFLAGS}` use in ParflowTest.cmake
+ Fixes MPIEXEC_PRE/POSTFLAGS passing in ParflowTest.cmake
  - Previously not being passed in quoted manner, which could be an issue when multiple, space-delimited flags are defined for use.
+ Fixes MPIEXEC_PRE/POSTFLAGS use in RunAmpsTest.cmake
  - Previously not using MPIEXEC_POSTFLAGS at all.
  - Previously not splitting MPIEXEC_PREFLAGS arguments  which could be an issue when multiple, space-delimited flags are defined for use.
+ Cosmetic quality-of-debugging changes to RunAmpsTest.cmake, RunParallelTest.cmake, and RunFileCompareTest.cmake.
  - Made to use have same output format
  - ~~Made to output the actual command being executed (using execute_procedure's COMMAND_ECHO argument)~~
    * Update (7/9/2020): COMMAND_ECHO is not an argument in the cmake version we support (3.14)
  - Made to output the command supposedly being executed.
  - Renames variable stdout to joined_stdout_stderr to be more correct.
+ Adds AMPS_TEST_FORCE_MPIEXEC to CMake option.
  - On some systems, it is necessary for any binary compiled with MPI to be executed with the appropriate ${mpiexec} command.
  - Setting AMPS_TEST_FORCE_MPIEXEC when configuring forces sequential AMPS tests to be executed with the ${MPIEXEC} command using 1 rank.
+ Bumps cmake minimum required version from 3.4 to 3.14 in RunAmpsTest.cmake, RunFileCompareTest.cmake, and RunParallelTest.cmake (which is the cmake version minimum version used in the main cmake file, [parflow/CMakeLists.txt](https://github.com/parflow/parflow/blob/master/CMakeLists.txt), which is also the only other location, besides what is included in these changes, where the cmake minimum version is defined).


smithsg84:
Thanks for patch, will merge.

PR #258, opened 2020-08-11
Clean clm slopes js

jswilley-csm:
Hi all,

The changes made in this fork will help to more accurately calculate energy fluxes at the ground's surface when using the coupled model CLM. Slope values are passed from ParFlow to CLM so that CLM can use them to account for slope and aspect when calculating solar incidence angle. 

Feel free to reach out with any questions to: jswilley@mymail.mines.edu

Best,
Jackson


jswilley-csm:
Hi all

I'm going to add a key to turn our new functionality on and off for the sake of backwards compatibility.

Best,
Jackson

PR #267, opened 2020-09-04
Parallel NetCDF dataset compression

sluehrs:
This pull request adds configurable deflate (zlib) compression capabilities to the NetCDF based writing routines. The possibilities of parallel data compression will only work in combination with the latest **NetCDF4 v4.7.4** release.

Two new configuration options can be used:

`pfset NetCDF.Compression True` Enable deflate based compression (default: `False`)
`pfset NetCDF.CompressionLevel 1` Compression level (`0-9`) (default: `1`)

This work was implemented as part of EoCoE-II project (www.eocoe.eu). 

Firsts benchmark tests show that the datasize for regular NetCDF output files could be lowered by a factor of three, in addition, as less data is written to disk, the reduced data size can also lower the overall I/O footprint towards the filesystem. Therefore, depending of the selected setup, the compression overhead can be balanced by reduced writing times.

The resulting NetCDF4 data files, which contain compressed datasets, are still readable by every normal NetCDF4 access mechanic without additional dependency.


smithsg84 (approved):

smithsg84:
Thanks for adding the compression flags.

PR #273, opened 2020-09-18
Fix venv in docker for pfpython install step

jourdain:


jourdain:
@smithsg84 small fix to support install step of pfpython which requires `python3-venv` inside docker

smithsg84 (approved):

PR #278, opened 2020-10-06
test refactor

grapp1:
Refactoring TCL and Python tests and changing some directory structure within parflow/test/.


grapp1:
@smithsg84 this PR is ready for review and merge.

test/python/CMakeLists.txt
```diff
@@ -1,15 +1,158 @@
...
+include(ParflowTest)
+include(ProcessorCount)
 
-add_subdirectory(base)
-add_subdirectory(base_2d)
-add_subdirectory(base_3d)
-add_subdirectory(silo)
-add_subdirectory(clm)
-add_subdirectory(washita/py_scripts)
+ProcessorCount(N)
```
smithsg84, line 4: This is not going to work.  The tests may be run over committed, might be running on a distributed machine so local count core count doesn't reflect number of MPI ranks available, and may be compiling on a front-end machine which has different core count from the compute nodes.

I get what the goal is but this will break most of several of the testing platforms including the builds on TravisCI.

Open to suggestions but this can't stay as it is.

smithsg84 (approved):
There is one issue that needs to get fixed at some point.   We can't assume that # of local cores is number of ranks available for testing.

test/python/CMakeLists.txt
grapp1: Noted. It wasn't working that well in our testing anyway. We'll remove it in the forthcoming PR.

PR #281, opened 2020-10-19
DRAFT: Py metadata

jourdain:
DO NOT MERGE - WIP


pf-keys/definitions/metadata.yaml
```diff
@@ -0,0 +1,310 @@
...
+      domains:
+        AnyString:
+
+    StopTime:
+      help: >
+        [Type: string] Ending time of CLM data (formatted in 'HH-MM-SS').
+      domains:
+        AnyString:
+
+    clm_ic:
```
jourdain, line 69: Should the name be `InitialCondition`?

pftools/python/parflow/tools/builders.py
```diff
@@ -705,47 +728,50 @@ def print_as_table(self, props_in_header=True, column_separator='  '):
...
-        self.output = {}
-        self.name_registration = {}
-        self.column_index = {}
-        self.props_in_row_header = True
-        self.table_comments = []
-        yaml_key_def = os.path.join(
-            os.path.dirname(__file__), 'ref/table_keys.yaml')
-        with open(yaml_key_def, 'r') as file:
-            self.definition = yaml.safe_load(file)
+        self.reference_file = 'ref/table_keys.yaml'
```
psavery, line 731: It might be a little cleaner if this is a property too (then you don't need to define it before calling `super()`.

jourdain:
Already taking care in another PR

PR #289, opened 2020-11-11
Python pftools 1.0.0

psavery:
- Added ``get_*_from_location`` methods to get attributes at
arbitrary locations. This includes details, docs, and value.
- Added ``from_definition`` class method to automatically create
a ``Run`` object from a file path.
- Added tests for both of the above features.


pftools/python/parflow/tools/core.py
```diff
@@ -223,6 +223,37 @@ def __init__(self, name, basescript=None):
...
+            yaml
+            pfidb
+
+        Args:
+            file_path (str): path to the file to read in
+
+        Returns:
+            A new Run object
+        """
+        file_path = Path(file_path)
```
jourdain, line 240: We need to support env variable like we do in our other path

pftools/python/parflow/tools/database/core.py
```diff
@@ -651,6 +648,52 @@ def _process_dynamic(self):
...
+        location = os.path.normpath(location)
+
+        split = location.split('/')
+        path, key = '/'.join(split[:-1]), split[-1]
+
+        if key == '.':
+            return self, None
+
+        parent = self
+        while key == '..' and getattr(parent, '_parent_', None):
```
jourdain, line 686: hum, it might be right, but does the key can be `..`. I thought those `..` will be at the head of the path...

pftools/python/parflow/tools/export.py
```diff
@@ -30,7 +30,7 @@ def _extract_sub_surface_props(self, geom_item):
         entry = {'key': name}
         has_data = False
         for key in self.pfkey_to_alias:
-            value = geom_item.get(key, skip_default=True)
+            value = geom_item.get_value_from_location(key, skip_default=True)
```
jourdain, line 33: Is the key notation from `pfkey_to_alias` using the `A.B.C` notation?

jourdain: Actually, it might be a single key and that is why it must be working.

jourdain: Actually that is not true, [see](https://github.com/parflow/parflow/blob/master/pftools/python/parflow/tools/ref/table_keys.yaml), we use the `.` notation

jourdain: The tests that validate that part must just print the table rather than comparing its actual content.

pftools/python/parflow/tools/database/core.py
psavery: So, because of the way I'm splitting the path, whatever is after the last `/` is considered the key. But it is not really a key if it is `..`.

This while loop is mostly to handle things like if the user put in `../..`. If they do that, the key will be `..`, which is not really a key.

For something like `Geom/Porosity/..`, that automatically gets turned into `Geom` from the `os.path.normpath` call.

For something like `../Porosity`, that `..` will get handled by `get_selection_from_location()` call.

pftools/python/parflow/tools/core.py
psavery: Do you think we should use `get_absolute_path()` so that it both expands the variables, and uses the `settings.WORKING_DIRECTORY` if the path is not absolute?

jourdain: yes

jourdain: Although the `pf_work_dir` could be confusing in that case since it is not set. It might be better to use env_replace + current python chdir.

psavery: This is resolved by https://github.com/parflow/parflow/pull/289/commits/05d15de1cf7fb71a5b765714e8058bdbe282c36c

jourdain:
LGTM

pftools/python/parflow/tools/export.py
psavery: This is now being tested, and was fixed.

jourdain:
@smithsg84 this should be ready for your review/merge approval.

smithsg84 (approved):

jourdain:
Thanks @smithsg84 !!!

Version 1.0.0 is a big step forward in term of API normalization with really cool features... 
Having it inside pf/master is awesome!

PR #292, opened 2020-11-25
print_van_Genuchten

ben-bou:
Stores the van Genuchten parameters `alpha`, `n`, `sres`, `ssat` in `problem_data`
and prints them in `solver_richards.c` if `print_subsurf_data` is set. Memory usage is increased by these four vectors which  store one value for every gridpoint. Performance should not be affected because the values are generated anyway.


reedmaxwell:
> Stores the van Genuchten parameters `alpha`, `n`, `sres`, `ssat` in `problem_data`
> and prints them in `solver_richards.c` if `print_subsurf_data` is set. Memory usage is increased by these four vectors which store one value for every gridpoint. Performance should not be affected because the values are generated anyway.

this is a nice feature, we should get this PR through.  Is it ready to go?

smithsg84:
Branch has conflicts that have to be resolved.

reedmaxwell:
> Branch has conflicts that have to be resolved.

thanks, yes, that should have been obvious : )

@ben-bou this is a great feature add, it would be nice to have the conflicts resolved and get this pushed

smithsg84:
Not sure what is going on with the changes, there are a lot of differences in file not related to this patch.

ben-bou:
> Not sure what is going on with the changes, there are a lot of differences in file not related to this patch.

@smithsg84 I believe the file changes are there because I branched of @hokkanen 's CUDA11 branch. None of the conflicting files are actually relevant for this PR. How do we proceed with this?

smithsg84:
The PR needs to have the non-related changes removed.

smithsg84:
Why is the reading of the new vectors necessary?   It looks to me like the pd_alphas/N/sres/ssat arrays are being computed in the problem_saturation.c loop after the VGInit method is called so the read values will be overwritten.

This is going to impact performance more than suggested.   Yes the values are calculated but this is dramatically increasing the memory writes in the loop.   Going from 1 write to 5.    Memory movement is more important than FLOPS these days.

smithsg84:
I have reworked this feature to avoid keeping the values in problem data and reassigning the vectors each time saturation is invoked.

smithsg84:
The feature in the branch was reworked by extending the PFModule API to include methods for outputing module state.   This saves reating new problem data state and the vector assignments are only done when the output is needed.

The van Genuchten printing was merge in #336 

Thanks for submitting the code it was helpful to base things on even if we didn't use it directly.

PR #300, opened 2021-01-21
Fix CUDA patchloop ival initialization bug

hokkanen:
Fix n_ival initialization bug which does not show up in ctests, but influences the results for a production case.


smithsg84 (approved):
Thanks for bugfix.

PR #303, opened 2021-01-29
Fix a few issues with pftools python CLM

psavery:
1. The map and parameter file names are now pulled from the "drv_clmin.dat" file
   before attempting to load them. Previously, they were assumed to be
   "drv_vegm.dat" and "drv_vegp.dat".

2. When checking if the driver data is set, it now checks if land covers have
   been set. Previously, it checked if the type of "Sand" was set, but that
   is now always set since it defaults to "Constant".

3. Boolean type is now checked in the to_native_type() function.

This also bumps the version to 1.1.1.

### DIFF
--- a/pftools/python/parflow/tools/builders.py
+++ b/pftools/python/parflow/tools/builders.py
@@ -1229,20 +1229,26 @@ class CLMImporter:
     def __init__(self, run):
         self.run = run
 
-    def files(self, input='drv_clmin.dat', map='drv_vegm.dat',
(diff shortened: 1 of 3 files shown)


jourdain:
Tested on LW run and it fixed the issue I was seeing. +2

jourdain:
@smithsg84 can be merged when you want.

The version 1.1.1 is already pushed on pypi.

smithsg84 (approved):

PR #304, opened 2021-01-29
Add data accessor

jourdain:
Good to go now...

### DIFF
--- a/pftools/python/parflow/tools/core.py
+++ b/pftools/python/parflow/tools/core.py
@@ -15,7 +15,7 @@ import argparse
 
 from . import settings
 from .fs import get_absolute_path
-from .io import write_dict
+from .io import write_dict, DataAccessor
 from .terminal import Symbols as TermSymbol
 
 from .database.generated import BaseRun
@@ -231,6 +231,7 @@ class Run(BaseRun):
     """
     def __init__(self, name, basescript=None):
         super().__init__(None)
+        self._accessor_ = None
         self._process_args_ = get_process_args()
         self.set_name(name)
         if basescript is not None:
@@ -292,6 +293,15 @@ class Run(BaseRun):
             new_name (str): New name for run
         """
         self._name_ = new_name
+
+    @property
+    def data_accessor(self):
+        """Return a DataAccessor to streamline access to
+        numpy array of the various field this run is linked to.
+        """
+        if self._accessor_ is None:
+            self._accessor_ = DataAccessor(self)
+        return self._accessor_
 
     def write(self, file_name=None, file_format='pfidb',
               working_directory=None):
--- a/pftools/python/parflow/tools/io.py
+++ b/pftools/python/parflow/tools/io.py
@@ -93,6 +93,7 @@ def load_patch_matrix_from_pfb_file(file_name, layer=None):
 def load_patch_matrix_from_image_file(file_name, color_to_patch=None,
                                       fall_back_id=0):
     import imageio
+
     im = imageio.imread(file_name)
     height, width, color = im.shape
     matrix = np.zeros((height, width), dtype=np.int16)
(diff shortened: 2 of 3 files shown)


jourdain:
@psavery can you double check what I wrote?
Do I need to run some formatter or else?

pftools/python/parflow/tools/io.py
```diff
@@ -453,3 +454,381 @@ def read_clm(file_name, type='clmin'):
...
+
+    @property
+    def times(self):
+        t0 = self._run.TimingInfo.StartCount
+        t_start = self._run.TimingInfo.StartTime
+        t_end = self._run.TimingInfo.StopTime
+        t_step = self._run.TimeStep.Value
+        t = t0 + t_start
+        time_values = []
+        keep_time_step = True
```
psavery, line 513: Looks like this is unused

pftools/python/parflow/tools/io.py
```diff
@@ -453,3 +454,381 @@ def read_clm(file_name, type='clmin'):
...
+    # ---------------------------------------------------------------------------
+    # Region selector
+    # ---------------------------------------------------------------------------
+
+    @property
+    def selector(self):
+        return self._selector
+
+    @selector.setter
+    def set_selector(self, selector):
```
psavery, line 553: This function might need to be named `selector`

pftools/python/parflow/tools/io.py
```diff
@@ -453,3 +454,381 @@ def read_clm(file_name, type='clmin'):
...
+        if press_type == 'PFBFile':
+            geom_name = self._run.ICPressure.GeomNames
+            if len(geom_name) > 1:
+                raise Exception(f'ICPressure.GeomNames are set to {geom_name}')
+            return self._pfb_to_array(self._run.Geom[geom_name[0]].ICPressure.FileName)
+        else:
+            # HydroStaticPatch, ... ?
+            raise Exception(f'Initial pressure of type {press_type} is not supported')
+
+        return None
```
psavery, line 625: If a function ends with no return statement, it returns `None` by default. So this `return None` is not necessary - but it could be useful for clarity. It's up to you. Also, just plain `return` is the same as `return None` as well.

pftools/python/parflow/tools/io.py
```diff
@@ -453,3 +454,381 @@ def read_clm(file_name, type='clmin'):
...
+    # ---------------------------------------------------------------------------
+
+    @property
+    def pressure_boundary_conditions(self):
+        # Extract all BC names (bc[{patch_name}__{cycle_name}] = value)
+        bc = {}
+        patch_names = []
+
+        # Handle patch names
+        if patch_name is not None:
```
psavery, line 636: Looks like `patch_name` is not defined

pftools/python/parflow/tools/io.py
```diff
@@ -453,3 +454,381 @@ def read_clm(file_name, type='clmin'):
...
+    def computed_permeability_x(self):
+        return self._pfb_to_array(f'{self._run.get_name()}.out.perm_x.pfb')
+
+    @property
+    def computed_permeability_y(self):
+        return self._pfb_to_array(f'{self._run.get_name()}.out.perm_y.pfb')
+
+    @property
+    def computed_permeability_z(self):
+        return self._pfb_to_array(f'{self._run.get_name()}.out.perm_z.pfb')
```
psavery, line 607: I think you can use `self._name` for these

pftools/python/parflow/tools/io.py
```diff
@@ -453,3 +454,381 @@ def read_clm(file_name, type='clmin'):
...
+        """Atmospheric Pressure [pa]"""
+        return self._clm_forcing('Press')
+
+    @property
+    def clm_forcing_spfh(self):
+        """Water-vapor specific humidity [kg/kg]"""
+        return self._clm_forcing('SPFH')
+
+    def _clm_map(self, root):
+        lat_root =
```
psavery, line 790: Looks like this line is incomplete

psavery:
> @psavery can you double check what I wrote?
> Do I need to run some formatter or else?

You can `pip install flake8` and run flake8 on the files to check for pep8 compliance and syntax errors.

jourdain:
@reedmaxwell I think I got all the changes/updates/fix you made into that one. We will test it more on our side to make sure all the Accessor are working as expected and then we should be able to merge it.


@lecondon and @reedmaxwell: What do you think? Are we missing any other accessor that will be useful?

asteroidb612:
@jourdain Would this be the place to put an accessor for elevation?

jourdain:
> @jourdain Would this be the place to put an accessor for elevation?

Yes, but what should be:
- Property name
- Property Defaults?
   - Type: Constant 
   - Value: 0 
   - FileName: {run_name}.DEM.pfb (Only in accessor, None otherwise in the keys)

@reedmaxwell @lecondon any idea? Does the following make sense to you?
```
ComputationalGrid.Elevation.Type = Constant | PFBFile 
ComputationalGrid.Elevation.Value = 0 (only when Type == Constant)
ComputationalGrid.Elevation.FileName = '{run_name}.DEM.pfb'
```

jourdain:
I think this is good to go @smithsg84

psavery (approved):
LGTM

pftools/python/parflow/tools/core.py
```diff
@@ -231,6 +231,7 @@ class Run(BaseRun):
     """
     def __init__(self, name, basescript=None):
         super().__init__(None)
+        self._accessor = None
```
psavery, line 234: Should this be named `_accessor_` with the extra underscore at the end so it is considered to be a private variable [here](https://github.com/parflow/parflow/blob/8f8703d20daafdf56f2d3082cd389886db3e00f1/pftools/python/parflow/tools/database/core.py#L67-L72) for the check [here](https://github.com/parflow/parflow/blob/8f8703d20daafdf56f2d3082cd389886db3e00f1/pftools/python/parflow/tools/database/core.py#L203)?

jourdain: Good catch... Thanks for finding that.

jourdain:
Working great now on the ML side of things. Thx @psavery to figure it out.

smithsg84 (approved):

Issue #317, opened 2021-03-22
'Unable to update metadata for' Output


lmthatch:
I'm currently running a simulation on NCAR's Cheyenne over 252 processors. Run parameters are adjusted every 24 hours and the simulation is restarted. Simulations appear to run fine, CLM and ParFlow logs are normal, however, in Cheyenne's error output file it repeats the line 'Unable to update metadata' for each of the 7 clm forcing parameters over 6000 times (I believe this is 24 hours x 252 processors). Previous simulations without daily restarts using the same forcing data did not produce this output.

```
Unable to update metadata for "downward shortwave radiation"
Unable to update metadata for "downward longwave radiation"
Unable to update metadata for "precipitation"
Unable to update metadata for "temperature"
Unable to update metadata for "wind velocity"
Unable to update metadata for "atmospheric pressure"
Unable to update metadata for "specific humidity"
```

I traced the output line back to the `MetadataUpdateForcingField` function in the `metadata.c` code, and it looks like the code is returning a zero when it encounters this particular issue after it puts out the noted output line, so I am unsure if this is actually an error or problem. The issue does not appear to be impacting model output (on first glance) and modeled SWE which I've found to be a good indicator of CLM restart problems appears normal in timing and volume.

@sluehrs Reed recommended I post this here and specifically reach out to the kitware team. Thanks!

jourdain:
Thanks @lmthatch for reporting such issue. Let see if @vibraphone has any good idea/solution for that part. Maybe an environment variable to disable any metadata print? Could be useful for big runs like that?

Otherwise the question could be: 
- Now that we have Python pftools, should the metadata handle at that level? Generate the .pfmetadata via some cli helper inside pftools? Or do we need something that only the simulation runtime can provide, in which case the .c approach is the way to go?

@reedmaxwell any comment, suggestion ?

Thanks!

reedmaxwell:
Thanks @jourdain for the quick reply.  It depends on what metadata you want to keep for long runs.  The timing (times / timesteps / date) and some CLM output / restart intervals may be the only things that have to come from ParFlow itself.  The timesteps could be predictable but if the solver fails it cuts the timestep in half and tries again so there's no way of really knowing prior to the run.  @lecondon thoughts on this?

vibraphone:
@lmthatch  A little more information would verify what I think is happening:
+ I assume CLM vegetation forcing was off for your run?
[...]

PR #318, opened 2021-03-23
Make errors less verbose and more descriptive.

vibraphone:
Do not print improper-metadata setup error messages
on every process at every timestep; print once at
initialization with a more complete description of
the problem.


vibraphone:
This should address #317 at least in part.

smithsg84 (approved):

PR #326, opened 2021-04-13
Hydrology module

callachennault:
* Added `hydrology.py` module to `pftools/python/parflow/tools`. Contains functions for calculating water table depth, subsurface storage, surface storage, evapotranspiration, and overland flow.
* Modified `DataAccessor` class in `pftools/python/parflow/tools/io.py` to contain properties/methods for each of the hydrology functions above. Also added properties for any values that the hydrology calculation functions require as parameters.
* Also in `pftools/python/parflow/tools/io.py`: Renamed the `_clm_forcing` function to `clm_forcing`. It would be useful to make this function public so that users can call this function using an argument vs calling the individual variants of the function. 
* Modified `tests/python/new_features/selection_methods.py` and the corresponding `dsingle.pfidb.ref` to test out accessing the `Cell.{nzListNumber}.dzScale.Value` attributes that we ran into an issue with.
* Modified `tests/python/new_features/write_check.py` to line up with the attributes added to `dsingle.pfidb.ref` for the above test.

Changes 4/15:
* In `pftools/python/parflow/tools/io.py`: Added `clm_output` method to `DataAccessor` class which can read a specified CLM output field into a Numpy array from a multi-layer `.C.pfb` or individual `.pfb` file. Also added `clm_output_variables` property, which contains the shortened names of the 14 CLM output variables.
* Also in `pftools/python/parflow/tools/io.py`: modified the `DataAcessor` `et` property to read ET data using the `clm_output()` function when CLM output is enabled.


callachennault:
@jourdain Can you review this? Do I need to modify the pypi version as well? Thanks!

reedmaxwell:
@smithsg84 can you also review?
@callachennault have these new commands been documented in the manual?

callachennault:
> @callachennault have these new commands been documented in the manual?

@reedmaxwell I haven't added any documentation to the manual, what section should I put it in?

jourdain (approved):
Looks great!

reedmaxwell:
> > @callachennault have these new commands been documented in the manual?
> 
> @reedmaxwell I haven't added any documentation to the manual, what section should I put it in?

@callachennault the PFTools part of the manual is [here](https://github.com/parflow/parflow/blob/master/docs/manuals/pftools.tex)

I am not sure the process for the read the docs, but the directory, etc, is [here](https://github.com/parflow/parflow/tree/master/docs/pf-keys)

@smithsg84 what are the doxygen recs?

jourdain (approved):
LGTM

callachennault:
Hi @smithsg84, can you review this PR? Thank you!

jourdain (approved):

smithsg84:
Please add an rst file in docs/pf-keys/tutorials.   Something like the domain_builder.rst file explaining basic usage for users.

Can a test be added for the new methods?   We really should have a test for regressions in the python functions.

Thanks

vineetbansal:
Hi @smithsg84 .. we do have some pytests (along with serialized .npy files against which we check the function outputs) against which all this code was validated, but for @callachennault  to read the existing structure of your tests and write something that conforms to their spirit will take a little while.

The same for documentation - the `DataAccessor` class for which we've added properties has existed for many months now, without any corresponding documentation, and we've just added additional properties here. So documenting existing code would have to come first before we can document what's relevant to this PR.

This is not to say that we're shying away from adding these two, but we do need the code merged in `pftools` to be able to use it in the very short run. Perhaps we create an *issue* on this, assign it to @callachennault or @vineetbansal  and we can submit 2 additional PRs for the docs/tests after mid-May (when our proposal is due?)

smithsg84 (approved):
Need testing and more documentation

smithsg84:
Created issue #329  to complete documentation and testing

PR #327, opened 2021-04-20
Add Kokkos support for accelerators

hokkanen:
This PR adds Kokkos accelerator backend as an alternative to the native ParFlow CUDA backend to support more accelerator devices. The implementation does not rely on any CUDA-specific arguments but still requires Unified Memory support from the accelerator devices. It should be compatible with AMD GPUs when sufficient Unified Memory support is available from this vendor.

The performance of using CUDA through the Kokkos library is slightly worse in comparison to the ParFlow native CUDA implementation. This is because a general Kokkos implementation cannot leverage certain CUDA features such as cudaMemset() for initializations or CUDA pinned host/device memory for MPI buffers. Also, Kokkos determines grid and block sizes for compute kernels differently.

The RMM pool allocator for Unified Memory can be used with Kokkos (when using Kokkos CUDA backend) and improves the performance very significantly. In the future, a Unified Memory pool allocator that supports AMD cards is likely needed to achieve good performance.

User instructions on how to use Kokkos backend can be found from [README-GPU.md](https://github.com/hokkanen/parflow/blob/kokkos/README-GPU.md). Basically, the following must be specified for CMake when compiling ParFlow:

```-DPARFLOW_ACCELERATOR_BACKEND=kokkos -DKokkos_ROOT=/path/to/Kokkos```

Furthermore, the performance of the simulation initialization phase has been improved significantly when using GPUs (with CUDA and Kokkos).

### CODE REVIEW

File: pfsimulator/parflow_lib/pf_omploops.h [hunk 1]
```diff
@@ -25,7 +25,7 @@
  * @brief OpenMP functions and loop definitions
  */
 
-#if PARFLOW_ACC_BACKEND != PARFLOW_BACKEND_OMP
+#ifdef PARFLOW_HAVE_OMP
```
smithsg84, line 28: I'm confused by this logic change.   Looks like it is turning off these OMP macros when OMP is present and attempting to include omp.h when HAVE_OMP is false?

pfsimulator/parflow_lib/overlandflow_eval_Kin.c
```diff
@@ -25,7 +25,7 @@
 
 #include "parflow.h"
 
-#if PARFLOW_ACC_BACKEND != PARFLOW_BACKEND_CUDA
+#if !defined(PARFLOW_HAVE_CUDA) && !defined(PARFLOW_HAVE_KOKKOS)
```
smithsg84, line 28: I thought the pattern being established was to  check PARFLOW_ACC_BACKEND in non-backend specific code and the HAVE_CUDA etc was used only in backend specific files.

It looks like we have multiple defines for the same thing.  What is the difference?

smithsg84 (commented):
Confused by one of the OMP #ifdef logic changes, please verify this is correct.
Thanks,
Steve

pfsimulator/parflow_lib/overlandflow_eval_Kin.c
hokkanen: I decided to remove the whole PARFLOW_ACC_BACKEND thing completely, because it felt problematic and just use PARFLOW_HAVE_XXXXX everywhere. Just one definition now. It became less readable when not having CUDA or KOKKOS for example. Or with some other combinations.

Doing that also allows using the following mixed combination of configure args for development purposes (helps profiling the performance differences between CUDA and Kokkos, not intended for users): 
```-DPARFLOW_ACCELERATOR_BACKEND=cuda -DKokkos_ROOT=/path/to/Kokkos``` 

Basically, with this option, cudaMemset is used for initializations, and Kokkos uses some CUDA-specific template arguments, eg, instead of calling architecture-agnostic
```Kokkos::kokkos_malloc<Kokkos::HostSpace>(size)```
the CUDA-specific host memory allocation is used which allocates page-locked host memory for faster host-device data transfer:
```Kokkos::kokkos_malloc<Kokkos::CudaHostPinnedSpace>(size)```

pfsimulator/parflow_lib/pf_omploops.h
hokkanen: This was a bug, fixed.

pfsimulator/parflow_lib/overlandflow_eval_Kin.c
smithsg84: OK, I understand now.  Thanks.

smithsg84 (approved):
Thanks for all the GPU work.

PR #328, opened 2021-04-20
Prevent eager CLM export when no clm keys have been defined

jourdain:
By default when CLM is used, we always try to export the .dat file
using the pf keys and pfb files. But if those were never defined,
such opperation would fail.

This fix try to guard against it.


jourdain:
This issue was discovered by Elena.

FYI @vineetbansal @reedmaxwell @lecondon

jourdain:
Elena and @vineetbansal tested it on their side and it solved their issue.

@smithsg84 when ever you feel like it, you can review/merge it.

Thanks!

smithsg84 (approved):

PR #346, opened 2021-06-14
Support for extra parameters to pfmask-to-pfsol

vineetbansal:
We've identified a need to use the `SolidFileBuilder` class to generate `.pfsol` files, but with a customized value of the `z-top` parameter (i.e. invoking the underlying `pfmask-to-pfsol` parflow utility with an additional `--z-top 1000` flag. This minor change should allow us to use the `SolidFileBuilder` class with any custom flags not exposed as explicit attributes in the class.

### DIFF
--- a/pftools/python/parflow/tools/builders.py
+++ b/pftools/python/parflow/tools/builders.py
@@ -104,7 +104,7 @@ class SolidFileBuilder:
         self.patch_ids_side = side_patch_ids
         return self
 
-    def write(self, name, xllcorner=0, yllcorner=0, cellsize=0, vtk=False):
+    def write(self, name, xllcorner=0, yllcorner=0, cellsize=0, vtk=False, extra=None):
         """Writing out pfsol file with optional output to vtk
 
         Args:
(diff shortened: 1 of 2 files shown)


pftools/python/parflow/tools/builders.py
```diff
@@ -104,14 +104,16 @@ def side_ids(self, side_patch_ids):
         self.patch_ids_side = side_patch_ids
         return self
 
-    def write(self, name, xllcorner=0, yllcorner=0, cellsize=0, vtk=False):
+    def write(self, name, xllcorner=0, yllcorner=0, cellsize=0, vtk=False, extra=None):
```
smithsg84, line 107: Is there a reason not to have the default value of extra = [] to simplify the logic slightly?

vineetbansal: This is a common pattern I've seen in Python for many years (avoid mutable default arguments) to avoid a situation where changes in one invocation of a function/method alter another completely independent invocation. For example:
```
def f(a=[]):
    a.append(42)
    return a

print(f())
print(f())
```
results in:
```
[42]
[42, 42]
```

This happens because default arguments are evaluated **once** when the function is defined, not each time the function is called. Even though this particular method (`write`) is not altering the `extra` argument, its encouraged to avoid having a list as a default argument at all to avoid this gotcha (most IDEs/linters will point this out), should one's implementation get more complex in the future.

smithsg84: Thanks for great explanation,  I'm still getting up-to-speed on python.

smithsg84 (approved):

PR #370, opened 2022-02-23
Remove reference to parflowio, relax requirement on numba

arbennett:
This PR fixes a leftover reference to parflowio that wasn't removed in #365. It also makes the numba dependency optional as there are still some issues with installing on Mac M1 based machines.

### DIFF
--- a/pftools/python/setup.py
+++ b/pftools/python/setup.py
@@ -23,20 +23,27 @@ setup(
     packages=find_packages(),
     install_requires=[
         'pyyaml==5.4',
-        'numpy',
-        'xarray',
-        'numba',
-        'dask',
-        'imageio>=2.9.0'
     ],
     include_package_data=True,
     extras_require={
         'all': [
-            'imageio>=2.9.0',
+            'imageio>=2.9.0'
+            'numpy',
+            'xarray',
+            'numba',
+            'dask',
         ],
         'pfsol': [
             'imageio>=2.9.0'
-        ]
+        ],
+        'io':[
+            'numpy',
+            'xarray',
+            'dask',
+        ],
+        'fastio': [
+            'numba',
+        ],
     },
     entry_points={
         'xarray.backends': [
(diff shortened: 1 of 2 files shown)


pftools/python/setup.py
```diff
@@ -23,20 +23,27 @@
     packages=find_packages(),
     install_requires=[
         'pyyaml==5.4',
```
jourdain, line 25: Maybe relax that constraint to be >=5.4

pftools/python/setup.py
```diff
@@ -23,20 +23,27 @@
...
-        'numba',
-        'dask',
-        'imageio>=2.9.0'
     ],
     include_package_data=True,
     extras_require={
         'all': [
-            'imageio>=2.9.0',
+            'imageio>=2.9.0'
+            'numpy',
```
jourdain, line 31: is numpy core? or can it be truly optional? I don't know, I'm just wondering.

jourdain (approved):
Your change looks great! Thanks for making them. Might be worth doing 2 edits.
1) Update the version in setup.py so we can publish a new one.
2) Relax the yaml version constraint

pftools/python/setup.py
arbennett: I also don't know. I think this is how it was previously.

jourdain (approved):
LGTM but make sure you rebase on master. As I think you are slightly off master.

jourdain:
If everything comeback green we should be able to merge. 

@callachennault do you want to try to publish the new version to PyPI just to make sure you are all-set on your side?

smithsg84 (approved):

callachennault:
> If everything comeback green we should be able to merge.
> 
> @callachennault do you want to try to publish the new version to PyPI just to make sure you are all-set on your side?

@jourdain I wasn't able to get it working. I don't know if I'm missing a prerequisite to the steps [here](https://github.com/parflow/parflow/tree/master/pftools/python#how-to-update-this-package-developers-only).

jourdain:
I just forwarded you an email explaining what I had to do. Hopefully that should help you.

PR #371, opened 2022-02-25
Numba fix, readme update, version number upgrade

callachennault:
Small changes to the latest PFTools release:

- Updates to the README to detail the PFTools package update process (@jourdain Let me know if it looks correct!)
- Fix to jit decorator for numba
- Version number upgrade to 1.3.2

### DIFF
--- a/pftools/python/parflow/tools/io.py
+++ b/pftools/python/parflow/tools/io.py
@@ -26,11 +26,11 @@ try:
     from numba import jit
 except ImportError:
     # Some systems may not have numba capabilities
-    def jit(self, function, *args, **kwargs):
+    def jit():
         """Dummy decorator, does nothing"""
-        def wrapper(*args, **kwargs):
-            function(*args, **kwargs)
-        return wrapper
+        def _decorator(func):
+            return func
+        return _decorator
 
 from numbers import Number
 import numpy as np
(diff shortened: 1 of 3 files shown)


pftools/python/parflow/tools/io.py
```diff
@@ -26,11 +26,11 @@
     from numba import jit
 except ImportError:
     # Some systems may not have numba capabilities
-    def jit(self, function, *args, **kwargs):
+    def jit():
```
jourdain, line 29: Maybe to be on the safe side `def jit(*args, **kwargs):`

jourdain (approved):

pftools/python/parflow/tools/io.py
callachennault: Sure, I could make that change. Should I do another release? 1.3.2 contains the version in this PR: `def jit():`

jourdain: No I think you are good for now. It is just needed if ever other use of the @jit() decorator is happening. That way we would not care if args or kwargs are passed. So far, the code is not passing anything, but things may change down the road.

jourdain:
LGTM

smithsg84 (approved):

PR #372, opened 2022-03-04
Python PFTools vegm file reader fix

callachennault:
The `_read_vegm` function in PFTools is inconsistent with parflow-python xy indexing. This fix swaps the indices so that the x index of the `vegm_array` correctly reflects the columns and y index reflects the rows.


jourdain (approved):
LGTM

smithsg84 (approved):

PR #378, opened 2022-03-24
Increased input database maximum value size from 4097 to 65536

alanquits:
### Update summary:

the maximum input database value length was increased from 4097 to 65536. A bounds check is performed that emits a helpful error message when a database value is too big.

4097 to 65536 seems like a big increase, but IDB_MAX_VALUE_LEN is only used once as a buffer size to read database entries. The char pointers are then null-terminated and passed to IDB_NewEntry, where strdup mallocs an appropriate string size. The strings are still freed in the same place.

### Checks performed:

1. All tests pass on my personal laptop configured with Hypre and CLM, except for those that exceed my available processors.
2. A modified copy of impes.simple.tcl was configured with a key of size 65535. It ran without segfaulting.
3. A modified copy of impes.simple.tcl was configured with a key of size 65536. It produced the desired error message in the output log:
_"Node 0: Error: The value associated with input database key 'A.Very.Long.Key' is too long. The maximum length is 65535. "_

### DIFF
--- a/pfsimulator/parflow_lib/input_database.c
+++ b/pfsimulator/parflow_lib/input_database.c
@@ -162,11 +162,11 @@ IDB *IDB_NewDB(char *filename)
     amps_SFBCast(amps_CommWorld, file, invoice);
     if ((value_len + 1) > IDB_MAX_VALUE_LEN) {
       key[key_len] = '\0';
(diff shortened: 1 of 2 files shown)


smithsg84 (approved):
Thanks for error checking.

PR #380, opened 2022-04-02
FIx errors in python input script for LW test case

aureliayang:
Fix errors in this file. Now it can run successfully and users can use more than one CPU core (both didn't work previously).


smithsg84 (approved):
Thanks for updating.

PR #383, opened 2022-04-06
vshape should be a 1d array instead of a 2d array.

kvrigor:
Its attributes are specified as `[INTEGER, DIMENSION(2*id var nodims(1)), IN]`
based on the [OASIS3-MCT docs](https://gitlab.com/cerfacs/oasis3-mct/-/blob/OASIS3-MCT_3.1/doc/oasis3mct_UserGuide.pdf):

```
• CALL oasis def var (var id, name, il part id, var nodims, kinout,
var type, kinfo) or
• CALL oasis def var (var id, name, il part id, var nodims, kinout,
var actual shape, var type, kinfo) or
• CALL prism def var proto(var id, name, il part id, var nodims, kinout,
var type, kinfo) or
• CALL prism def var proto(var id, name, il part id, var nodims, kinout,
var actual shape, var type, kinfo)
.
.
.
– var actual shape [INTEGER, DIMENSION(2*id var nodims(1)), IN]: is not
used anymore. The interface has recently been overloaded, and this argument is no longer re-
quired. But for backwards compatibility, it can still be passed; if so, it has to be a vector of
integers of any length (for simplicity we advise to pass a vector of length 1).
```

### DIFF
--- a/pfsimulator/amps/oas3/oas_pfl_define.F90
+++ b/pfsimulator/amps/oas3/oas_pfl_define.F90
@@ -59,7 +59,7 @@ REAL (KIND=8), INTENT(IN)                  ::                   &
 CHARACTER(len=4)                           ::  clgrd
 INTEGER                                    ::  il_flag            ! Flag for grid writing by proc 0
 INTEGER                                    ::  var_nodims(2)      ! used in prism_def_var_proto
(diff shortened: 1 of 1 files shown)


smithsg84:
Thanks for update.

PR #385, opened 2022-04-08
DRAFT: Initial changes for ParFlow-eCLM coupling

kvrigor:
Hi Steven,

I maintain the [eCLM] land surface model and our team would like to run it with
ParFlow. I'm raising this PR to lay the foundation for my next set of PRs related
to eCLM+ParFlow coupling.This feature is not yet complete, but I think it's best
to hear your feedback before pursuing larger changes in the ParFlow codebase.

We want to modify the basic Richards' equation (eq. 5.3 in ParFlow manual) by
scaling the hydraulic conductivity tensor by an  _ice impedance_  factor calculated
from eCLM. _Ice impedance_ is a time-dependent quantity which represents the
increased tortuosity of water for ice-filled pore spaces (see eq. 2.7.47 [here]).
To accomplish this, we have to add ice-related args to [`KinsolNonlinSolver()`].

The key change in this PR is the new solver **RichardsCoupled** with its associated
code in `solver_richards_coupled.c`. Initially I implemented my changes in
`solver_richards.c`, but I finally decided to create a new solver to avoid breaking
backwards compatibility that may result from these next big changes:

* Restructuring of args of `CALL_send_fld2_clm()` and `CALL_receive_fld2_clm()`
  to accommodate extra fields sent to/received from eCLM
* Adding ice parameters to [`*nonlin_solver`].

`solver_richards_coupled.c` does not contain the integrated CLM codes (wrapped
inside `HAVE_CLM`) like in `solver_richards.c`. Its logic will focus only on the
required "slight tweaks" of Richards' equation that would make eCLM+ParFlow
functional. I also think creating a solver specific to eCLM+ParFlow avoids further 
complicating `solver_richards.c` and thus makes code maintenance easier in the long run.

Is this the right approach? I'll be glad to hear your questions/comments/suggestions.

[eCLM]: https://github.com/HPSCTerrSys/eCLM
[here]: https://escomp.github.io/ctsm-docs/versions/release-clm5.0/html/tech_note/Hydrology/CLM50_Tech_Note_Hydrology.html#hydraulic-properties
[`KinsolNonlinSolver()`]: https://github.com/parflow/parflow/blob/32db1dc/pfsimulator/parflow_lib/kinsol_nonlin_solver.c#L200
[`*nonlin_solver`]: https://github.com/parflow/parflow/blob/1878384876ff2c28cd686a9b1398042af98d2708/pfsimulator/parflow_lib/solver_richards.c#L2769

Best Regards,
Paul Rigor


smithsg84:
Hi,

I'll have to look at this more carefully but some initial high level comments after a quick look.     I changed the PR to label it DRAFT: while we discuss the approach.   This seems a common thing on other OS projects.   DRAFT is remove when the PR is deemed ready for integration.

Overall I think this looks approach looks reasonable.   

Random stuff:

1) What is the reason for only being able to have CLM and eCLM be compiled in?   Is this a name conflict?

In general I think it is more user friendly if we have runtime selection rather than compile time selection.   Since we have control over the PF CLM we can probably fix function naming conflicts if that is the problem.

2) Extracting solver common code?  

We duplicate lots of code already in some parts of PF and I'd like to reduce/prevent that.   Looking at solver_richards and solver_richards_coupled there looks to be a lot of duplication (I/O setup).   Do you think we could refactor some of the common code between them to some shared functions?   (Not asking you to do all that work now but wondering if that is reasonable/possible since you have looked the code).

3) We should look at adding an OASIS build to the test suite.   

Currently it is not getting tested in the CI builds and if we have eCLM as an option we should be testing against it.

4) solver_richards/solver_richards_coupled

My biggest question is this split.  I need to look at code more closely to understand the trade-offs with having a single one with more internal options vs separated.   

I'll likely have more questions as I dig deeper.

Steve

kvrigor:
Sorry for the late reply! Here are my answers to your questions:

1. This behavior is consistent with `solver_richards.c` which contains two CLM
implementations: internal (`#ifdef HAVE_CLM`) and coupled (`#ifdef HAVE_OAS3`).
Enabling both CLMs at compile time gives compilation errors. Choosing CLM at
runtime would have been possible if its implementations were cleanly separated
in code; however, even if this is made possible, it would introduce another issue
of adding required dependencies (i.e. OASIS3-MCT and netCDF) to the build process.
I think these dependencies should only be provided when using coupled CLM. Selecting
CLM at compile time keeps the build process simple for ParFlow users which mostly
use only internal CLM.

2. I would also love to refactor code to reduce redundancy, but I think it's too
early to do it since it's not yet clear which codes should be shared/not shared
between `solver_richards.c` and `solver_richards_coupled.c`. This would 
become more apparent once ParFlow+eCLM becomes functional.

3. I'll add this.

4. Last week I've discussed this with my supervisor @skollet, and he's also not
entirely  convinced with the split. I did it to hopefully avoid these problems:
  * Mixing different CLM implementations in a single solver (see my answer to the
    1st question). 
  * Breaking backwards compatibility that may result from these next big changes:
    - Restructuring of args of `CALL_send_fld2_clm()` and `CALL_receive_fld2_clm()`
    to accommodate extra fields sent to/received from eCLM
    - Adding ice parameters to [`*nonlin_solver`].

I will be pushing more commits to demonstrate and clarify the need for
`solver_richards_coupled.c`. I hope this PR could remain open while I continue
work on this.

[`*nonlin_solver`]: https://github.com/parflow/parflow/blob/1878384876ff2c28cd686a9b1398042af98d2708/pfsimulator/parflow_lib/solver_richards.c#L2769

alanquits:
I have a third opinion. An underlying issue is that variables contained in solver_richards.c should be split into separate modules so it’s not as much of a deal If this additional  solver is split into a new file or not.

I am currently doing some work to implement a different forcing input method and I am thinking about taking the other forcing methods out of solver_richards after that too. Adding this new source file would duplicate some of my work. It’s not an insurmountable amount of additional work and I certainly wouldn’t advocate not adding a new feature because I *might * do something. I am only giving this opinion for consideration because it’s a concrete example of how duplicating this kind of code makes other kinds of work more time consuming.

kvrigor:
> Currently, if there is a lot of overlap between the existing and new richards solvers, it means future work on features supporting both may also need to be duplicated.

@smithsg84 and @skollet are also concerned about this. I'll modify my approach to just using `solver_richards.c` and raise a new pull request.

reedmaxwell:
Thanks, I also have concerns about this and would not have supported two versions of `solver_richards.c`.  Many of us developers have worked on these aspects of ParFlow for decades, please keep us in the loop on your long term plans so that we can provide insight and guidance moving forward.

PR #392, opened 2022-05-16
Improve xarray IO reading for multiple pfb types

arbennett:
This PR fixes some cases where PFB files with different z-dimension shapes could not be merged together in xarray. Notably this happens for surface parameters which have shape (1, ny, nx) which really should be represented by xarray by squeezing out the `z` dimension. This now happens in xarray transparently. Loading files with the standard `read_pfb` or `read_pfb_sequence` will not auto-squeeze dimensions.

This also includes a minor set of performance improvements on the reader where 1) subgrid headers can be memmapped rather than using the `struct` library 2) when reading a sequence of PFB files only the first file's subgrid header is loaded.

This also re-adds the use of `delayed`, which I discovered was not originally the cause of threadlocking. This reversion improves parallelism.


arbennett:
This is now ready for review and merge if approved.

arbennett:
@smithsg84 yep, that makes sense - thanks!

smithsg84 (approved):

PR #393, opened 2022-05-19
Wells table reader and dev workflow improvements

westb2:
Branch name is a misnomer. This PR includes functionality for:

1) A wells table reader + some table reader enhancements
2) Some dev workflow improvements for the docker build that take my build time from ~4:00 to <30 seconds
3) Adds some functionality to support writing similar assertions for the python versions of the TCL test cases

This is a bit more than I would ideally like in a single pull request but I wasn't expecting to need to do all this when I started and it's become a bit entangled now. If anything needs to be pulled out to be reviewed separately just let me know and I'm glad to do so :)


pftools/python/parflow/tools/builders.py
```diff
@@ -311,6 +311,10 @@ def __init__(self, run=None):
         self.column_index = {}
         self.props_in_row_header = True
         self.table_comments = []
+        self.empty_value_tokens = {
+            "-",
+            ""
+        }
```
westb2, line 317: Make it so that the empty string also counts as an empty value. This supports the use case of leaving cells blank in a csv

test/python/default_richards_wells.py
```diff
@@ -349,6 +351,25 @@
...
+
+perm_y_pfb = f"/{run_name}.out.perm_y.pfb"
+assert pfbs_are_equal_to_n_sig_figs(new_output_dir_name + perm_x_pfb, correct_output_dir_name + perm_x_pfb)
+
+perm_z_pfb = f"/{run_name}.out.perm_z.pfb"
+assert pfbs_are_equal_to_n_sig_figs(new_output_dir_name + perm_z_pfb, correct_output_dir_name + perm_z_pfb)
+
+for i in range(0,5):
+    press_pfb = f"/{run_name}.out.press.0000{i}.pfb"
+    assert pfbs_are_equal_to_n_sig_figs(new_output_dir_name + press_pfb, correct_output_dir_name + press_pfb)
```
westb2, line 371: Added the same assertions as in the TCL test cas

test/python/utils.py
```diff
@@ -0,0 +1,15 @@
+import xarray as xr
+from math import floor, log10
+
+
+def pfbs_are_equal_to_n_sig_figs(pfb1, pfb2):
```
westb2, line 5: This function support writing the same type of assertion as the TCL where we check that the run output is equal to the expected value within some tolerance

pftools/python/parflow/tools/builders.py
```diff
@@ -603,6 +607,14 @@ def apply(self, run=None, name_registration=True):
...
             self.run = run
 
+        # new names are user specified names such as s1 s2 s3 for soil permeability
+        if infer_key_names and name_registration:
+            new_names = ""
+            for name in self.output:
+                if isinstance(self.output[name], dict):
+                    new_names = new_names + name + " "
+            self.key_root.Names = new_names
+
```
westb2, line 617: The above lines support putting the names in the table and not needing to declare them separately in the run file

westb2: I've added a flag infer_key_names just in case there are unintended consequences I didn't catch. Flag is default set to false so this shouldn't break backwards compatibility

docker/CMakeLists.txt
```diff
@@ -13,6 +13,13 @@ add_custom_target(DockerBuildDevelopment
   VERBATIM
 )
 
+add_custom_target(DockerRunDevelopment
+        docker run --rm -it --workdir /home/ubuntu/parflow/src --env=DISPLAY --name=parflow-development -e LANG=en_US.utf8 -v ${CMAKE_CURRENT_SOURCE_DIR}/..:/home/ubuntu/parflow/src parflow-development
+        COMMENT
+        "Start docker image 'parflow-development' mounted to source directory to allow live rebuilding"
+        VERBATIM
+        )
+
```
westb2, line 22: This creates a make command "make DockerRunDevelopment" that when run starts the dev container and mounts it to your code so that you can make edits to your local code and update the container by running the "rebuild" command inside of it

docker/dev/Dockerfile
```diff
@@ -58,6 +58,13 @@ RUN apt-get update && \
 RUN groupadd -g 1000 ubuntu && \
     useradd -u 1000 -g ubuntu -m -N -s /bin/bash ubuntu
 
+# We are adding sudo capabilities to the ubuntu user so that devs can run the cmake build inside the container.
+# If we instead have devs use the root user when they execute parflow you get MPI errors
+RUN apt-get update \
+    && apt-get install -y sudo
+RUN adduser ubuntu sudo
+RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+
```
westb2, line 67: I know it is a weird thing to do to add sudo to a docker container but I think we need to because
1) You cannot run parflow as root inside the docker container or MPI will throw errors
2) You need root privileges to rebuild the code base within the container
So if we want to support people being able to iteratively rebuild within the container and also using it to execute parflow on test cases, they need to be logged in as ubunutu but also have root privelages

docker/dev/Dockerfile
```diff
@@ -192,7 +200,16 @@ RUN $CMAKE --build $BASE_PATH/parflow/build
...
 USER root
 RUN $CMAKE --install $BASE_PATH/parflow/build --prefix /opt/parflow
+
+# Install the locally built version of pftools
 USER ubuntu
+RUN python3 -m pip install /home/ubuntu/parflow/build/pftools/python
+RUN cp -R /home/ubuntu/parflow/build/pftools/python/parflow/tools/ref ~/.local/lib/python3.8/site-packages/parflow/tools/ref
+
+RUN rm -R /home/ubuntu/parflow/src
+RUN mkdir /home/ubuntu/parflow/src
```
westb2, line 210: pftools installation was previously not being finished

docker/dev/Dockerfile
```diff
@@ -192,7 +200,16 @@ RUN $CMAKE --build $BASE_PATH/parflow/build
...
 
 ENV PARFLOW_DIR /opt/parflow/
 WORKDIR /opt/parflow/
+
+RUN echo 'alias rebuild="~/parflow/src/docker/dev/rebuild.sh"' >> ~/.bashrc
```
westb2, line 215: enable the rebuild command

docker/dev/rebuild.sh
```diff
@@ -0,0 +1,4 @@
+$CMAKE --build $BASE_PATH/parflow/build
+sudo $CMAKE --install $BASE_PATH/parflow/build
+python3 -m pip install /home/ubuntu/parflow/build/pftools/python
+cp -R /home/ubuntu/parflow/build/pftools/python/parflow/tools/ref ~/.local/lib/python3.8/site-packages/parflow/tools/ref
```
westb2, line 4: This becomes the rebuild command within the dev container

pftools/python/parflow/tools/ref/well_keys.yaml
```diff
@@ -0,0 +1,198 @@
+# -----------------------------------------------------------------------------
```
westb2, line 1: Right now we can't support all the well keys because some of the subkeys such as the cycle are dynamically named. Talked with Reed and he said these need to be reworked anyways so to get in what is supported

westb2:
@lecondon @reedmaxwell @smithsg84 If any of you could review or have ideas of who would be correct to review this I would greatly appreciate it. Also, someone needs to approve the automated checks

smithsg84:
@westb2 I will schedule time to review this after the PF short-course, early next week.   To error on the side of caution, I don't want to disrupt the code just before a bunch of new users start trying to build.

smithsg84 (changes requested):
Thanks for updates; we desperately need to have the results checks in the python.   The current python tests are basically useless without this so super happy to see that going in.

There are few minor things that need to be fixed.   

Steve

PR #401, opened 2022-06-11
Add CI test for PARFLOW_AMPS_LAYER="oas3"

kvrigor:
This test verifies that the `HAVE_OAS3` section of `solver_richards.c` has no
compile errors. All unit tests succeed except those that use Richards solver.
Passing the failed tests would require running coupled ParFlow+CLM (not yet
supported by this CI workflow), so we simply exclude these tests.


misc/build_scripts/build.oasis3.ubuntu20.04
```diff
@@ -0,0 +1,77 @@
+#
```
smithsg84, line 1: Don't really like the location of this file but not sure where to put it.   The misc directory looks like it is mostly old stuff that should be removed.

Wondering if we should have a directory for CI related files, maybe test/ci ?  Since it is related to ci/testing?

Fine for now but just FYI I'm creating a ticket to cleanup misc so may decide to move this file.

smithsg84 (approved):
Thanks for MR.

Do the tests not run under GitHub actions?    Never mind, I just noticed the ctests are running a few things, went so quick I thought no PF tests were running.

Steve

misc/build_scripts/build.oasis3.ubuntu20.04
kvrigor: > Don't really like the location of this file but not sure where to put it.

I agree. I'll move this file the next time when I update the OASIS CI workflow.

Issue #402, opened 2022-06-18
Python PFTools incorrectly reads velocity subgrid info


zperzan:
The Python PFTools package incorrectly calculates subgrid information for velocity files, depending on the process topology. As an example, run default richards with multiple processors in the _x_ and _y_ directions:

```bash
tclsh default_single.tcl 2 2 1
```

Then try to read the one of the velocity files that was output:

```python
from parflow import read_pfb
velx = read_pfb('default_richards.out.velx.00001.pfb')
print(velx.min(), velx.max())
```

The result should be velocities many orders of magnitude out of the range of velocity values for this problem (for me, 1e-302 and 1e304). This is because the subgrid info and offsets are incorrect when set via `precalculate_subgrid_info`.

https://github.com/parflow/parflow/blob/3aceb0b927de725649953823c505f376d39564eb/pftools/python/parflow/tools/io.py#L877

PR #403, opened 2022-06-18
Python PFTools, read subgrid info from file

zperzan:
This addresses https://github.com/parflow/parflow/issues/402 

It adds the `read_subgrid_info` method to the `ParflowBinaryReader` class, which directly reads the subgrid header from file, rather than calculating that info using _nx_, _ny_, _nz_, _p_, _q_ and _r_. This is useful when reading velocity files. For example, since velx files report values at cell faces in the _x_ direction, if _p_ > 1 in the process topology, each subgrid includes one extra row/column in the _x_ direction. 

To fix the use case in https://github.com/parflow/parflow/issues/402, use `read_sginfo=True` when reading velocity files:

```python
from parflow import read_pfb

velx = read_pfb('default_richards.out.velx.00001.pfb',
                read_sginfo=True)
print(velx.min(), velx.max())
```

### DIFF
--- a/pftools/python/parflow/tools/io.py
+++ b/pftools/python/parflow/tools/io.py
@@ -56,7 +56,7 @@ except ImportError:
 
 
 def read_pfb(file: str, keys: dict=None, mode: str='full', z_first: bool=True,
-             read_sg_info: bool=False):
+             read_sg_info: bool=True):
     """
     Read a single pfb file, and return the data therein
 
--- a/pftools/python/parflow/tools/io.py
+++ b/pftools/python/parflow/tools/io.py
@@ -56,7 +56,7 @@ except ImportError:
 
 
 def read_pfb(file: str, keys: dict=None, mode: str='full', z_first: bool=True,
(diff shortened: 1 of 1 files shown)


pftools/python/parflow/tools/io.py
```diff
@@ -55,7 +55,8 @@ def _decorator(func):
     from yaml import Dumper as YAMLDumper
 
 
-def read_pfb(file: str, keys: dict=None, mode: str='full', z_first: bool=True):
+def read_pfb(file: str, keys: dict=None, mode: str='full', z_first: bool=True,
+             read_sg_info: bool=False):
```
smithsg84, line 59: The default behavior should be correctness so default for read_sg_info should be True.   Correctness first and optimizations as options, especially in cases if it fails to work correctly.

smithsg84 (changes requested):
I think the default should be in favor of correctness not speed that sometimes fails to work.

smithsg84 (approved):
Thanks for patch.

PR #405, opened 2022-06-27
DRAFT: Fix bug in calcluating subgrid numbers when not provided to reader

arbennett:
There was an off-by-one error that could occur when calculating the p,q,r values from the header and first subgrid header. This is fixed by checking if adding 1 to the number of subgrids in any of the dimensions would reduce the remainder.

Update: This should not yet be merged, I think I have found a secondary corner case where this causes new failures.


arbennett:
This is no longer needed with the more robust subgrid information reading implemented in #403.

PR #411, opened 2022-09-11
Clm update

reedmaxwell:
Fixed an issue that was identified by @danielletijerina  where some bare soil on vegetated surfaces wasn't being beta-limited in CLM.  Fixes to clm_thermal.F90 were implemented.  At the same time, CLM snow additions and dew corrections by LBearup were added.  A snow-age fix for deep snow was implemented along with canopy dew.  Suggest that @lecondon review changes to CLM.

A test case was added in a subdirectory to /tests/tcl/clm that necessitated a changed `pftest_clm.tcl` 
that looks for the correct output in one directory up.  This test hasn't been added to the test suite, I don't know how to add it, but it would be great to include it.


smithsg84:
I have updated to include the test in the test suite.

Cleaned up the pftest.tcl to enable setting the regression correct output directory so the pftest.tcl scripts can be used by CLM tests.   pftestFile now has an optional parameter to specify the directory.   So it can be invoked like:

```
set correct_output_dir "../../correct_output/clm_output"
pftestFile clm.out.press.$i_string.pfb "Max difference in Pressure for timestep $i_string" $sig_digits $correct_output_dir
```

lecondon (commented):
I have reviewed the clm changes and don't have any comments

reedmaxwell:
Thanks @smithsg84 thanks @lecondon !

PR #416, opened 2022-09-28
Feature surface pressure

reedmaxwell:
This update includes a check for high pressures within the domain, usually created by irregularities in the slopes.  Pressures above a threshold are flagged, logged and can be reset to a different pressure.


reedmaxwell:
in addition to code changes, a regression test has been added and the new keys are included in the read-the-docs manual

pfsimulator/parflow_lib/solver_richards.c
```diff
@@ -2885,6 +2888,69 @@ AdvanceRichards(PFModule * this_module, double start_time,      /* Starting time
...
+          // printf(" pp[ip] %10.3f \n",pp[ip]);
+          // printf(" NZ: %d \n",nz);
+          if (k == (nz - 1))
+          {
+            //   printf(" %d %d %d %d  \n",i,j,k,ip);
+            //   printf(" pp[ip] %10.3f \n",pp[ip]);
+
+            if (pp_sp[ip] > public_xtra->threshold_pressure)
+            {
+              printf(" pressure reset: %d %d %d %10.3f \n", i, j, k,
```
smithsg84, line 2941: One of the recommend practices is to not have output to stdout.   Use amps_Printf instead of printf so we can control output destination.

Is the current timestep useful in this diagnostic message?

pfsimulator/parflow_lib/solver_richards.c
```diff
@@ -2885,6 +2888,69 @@ AdvanceRichards(PFModule * this_module, double start_time,      /* Starting time
...
+                     pp_sp[ip]); pp_sp[ip] = public_xtra->reset_pressure;
+            }
+          }
+        }
+                     );
+      }
+    }
+
+    /* update pressure,  not sure if we need to do this but we might if pressures are reset along processor edges RMM */
+    handle = InitVectorUpdate(instance_xtra->pressure, VectorUpdateAll);
```
smithsg84, line 2951: I think the boundary update is only needed when reset_surface_pressure=1, so should be in the if {} block.

pfsimulator/parflow_lib/solver_richards.c
reedmaxwell: Thanks @smithsg84 that is a very good catch, please do make this change.  And yes, the current timestep info would be really helpful here.

pfsimulator/parflow_lib/solver_richards.c
reedmaxwell: ahh good point

reedmaxwell: @smithsg84 - quick Q: to fix this you just mean to move the VectorUpdate inside the if{surface_pressure} block?

smithsg84: Yes I think it can be moved there since we are adding for case when the threshold loop could change a value we only need to do the loop when the flag is set.

(I thought about a boolean to track if a value had been changed and only do the exchange if a threshold was found but duh that doesn't work since we don't know if changes are being made on other MPI tasks so that won't work.   I suppose we could do a reduction to see if anything changed on any task before doing the exchange.   That is likely faster if most cases don't have any thresholding.    That is a bit more complicated.    So much for quick answer to quick question :)

pftools/python/setup.py
```diff
@@ -7,7 +7,7 @@
 
 setup(
     name='pftools',
-    version="1.3.6",
+    version="1.3.6.1",
```
jourdain, line 10: This is an invalid version number. You need to go to `1.3.7` or even `1.4.0` since it is a new feature rather than a fix.

jourdain: Actually reading the spec, it should not matter.

pfsimulator/parflow_lib/solver_richards.c
reedmaxwell: thanks @smithsg84 I pushed a revised version and I think we are all set now

pftools/python/setup.py
reedmaxwell: thanks @jourdain fixed

reedmaxwell:
all requested changes should be in the latest update, I think we are GTG

smithsg84 (approved):

PR #426, opened 2022-10-24
Correct formula in OverlandSpinupDampPX description

jsacerot:
Based on http://parflow.blogspot.com/2015/08/spinning-up-watershed-model.html


smithsg84 (approved):
Thanks for the manual fix.

PR #428, opened 2022-10-28
Error in Eq. 4.35 - Subsurface storage

jsacerot:
Based on the [PFtools water balance formulation](https://github.com/parflow/parflow/blob/master/pftools/water_balance.c), it seems that (\psi) is repeated in the second half of Eq. 4.35. Also notice that ∆𝑥, ∆𝑦, and ∆z are not defined in Section 4 (MODEL EQUATIONS).

### DIFF
--- a/docs/user_manual/models.rst
+++ b/docs/user_manual/models.rst
@@ -750,8 +750,8 @@ being an array of balances over the domain) as follows:
    :label: sub_store
 
    \begin{aligned}
-   Vol_{subsurface} = \sum_\Omega [ S(\psi)S_s \psi \Delta x \Delta y \Delta z +
-   S(\psi)(\psi)\phi \Delta x \Delta y \Delta z]
+   Vol_{subsurface} = \sum_\Omega [ S(\psi) S_s \psi \Delta x \Delta y \Delta z +
+   S(\psi) \phi \Delta x \Delta y \Delta z]
    \end{aligned} 
 
 The surface storage is calculated over the upper surface boundary 


smithsg84 (approved):
Thanks for manual fixes.

jsacerot:
@smithsg84 as soon as you merged the commit, an issue appeared in the equations numbering in the [PDF version](https://github.com/parflow/parflow/issues/430) of the manual.

smithsg84:
This issue is present in in the Sept 6th generated PDF and is not related to this merge.

https://parflow.readthedocs.io/_/downloads/en/v3.11.0/pdf/

reedmaxwell:
I also don't see it in the current readthedocs version of the manual.  The PDF is generated from the RST files and my check of the most recent version of the [RTD](https://parflow.readthedocs.io/en/latest/models.html) and [PDF](https://parflow.readthedocs.io/_/downloads/en/latest/pdf/) appear to agree.  Am I missing something in this original comment?

smithsg84:
Yes it is present in latest versions but this issue was around before the #428 merge (it appears in PDF from Sept 6th) was done so I don't think this is related to issue #428.   I don't understand the "as soon as merged the commit" comment above.  What does this have to do with this merge?

jsacerot:
> Yes it is present in latest versions but this issue was around before the #428 merge (it appears in PDF from Sept 6th) was done so I don't think this is related to issue #428. I don't understand the "as soon as merged the commit" comment above. What does this have to do with this merge?

I thought the errors in equation numbering appeared after the changes proposed in this issue were merged into the master. Sorry for the confusion.

PR #435, opened 2022-11-10
Update ci images

smithsg84:
Update NetCDF used in the CI suite due to compilation issues with previous version and current GCC compilers.
Update image used in the CI suite for the non-CUDA/OASIS3 builds; 18.03 image is deprecated on GitHub.
Fix warning messages coming from GCC 11.3.0 and optimized builds.
Updated deprecated mallinfo command.


smithsg84:
PR has checks that do not exist anymore due to CI updates.

PR #441, opened 2022-12-02
Valgrind cleanup

smithsg84:
Cleanup memory leaks reported by valgrind.   This commit runs test suite with valgrind enabled.
Add a parflow.settings file to record information about the build.  Fix CMAKE warnings and improve NetCDF find macro to use pkg-config and CMake build info if available in the NetCDF install.


smithsg84:
Expected test list is out of date so some tests not running.

PR #459, opened 2023-02-21
Feature predictor

reedmaxwell:
This PR creates a feature to allow prediction of cells w/ pressures that switch from negative to positive pressures during a timestep, potentially reducing the number of solver iterations, particularly the function evaluations.  Keys are added to the database and RST files, feature defaults as false so baseline functionality is unchanged and backward compatibility is preserved.

New Keys Are
```
<runname>.Solver.SurfacePredictor = True [False]
<runname>.Solver.SurfacePredictor.PrintValues = True [False]
<runname>.Solver.SurfacePredictor.PressureValue = [0.00001]
```


reedmaxwell:
Also, since this feature adds new keys (like any PF feature would), this PR requires an update to the Python PFTools library to function.  Users must either go through the very painful and complicated process of rebuilding PFTools-python locally and installing an update python library from there, or we should update the PFTools on PyPi with this PR to make sure functionality is preserved.

smithsg84 (approved):

PR #462, opened 2023-02-28
Update hydrology.py

elena-leo:
In the OverlandKinematic boundary condition, the cells just outside the domain are used for the computation of flow into-out of the domain. This is the reason behind fixing the slopes outside eastern and northern edges. Similarly, also Manning's coefficient values need to be overwritten in those cells.


reedmaxwell (approved):
looks good to me, thanks for catching this!

smithsg84 (approved):
Thanks for patch.

PR #466, opened 2023-03-20
Error in Haverkamp saturation formula

jsacerot:
Typo in Haverkamp saturation formula: alpha replaced with A.

Correct formula from:

Haverkamp, R. and Vauclin, M. (1981), A Comparative Study of Three Forms of the Richard Equation Used for Predicting One-Dimensional Infiltration in Unsaturated Soil. Soil Science Society of America Journal, 45: 13-20. https://doi.org/10.2136/sssaj1981.03615995004500010003x


smithsg84 (approved):
Thanks for the documentation fix.

PR #472, opened 2023-06-18
Dirichlet bugfix

kvrigor:
`test_XYZTPlus1.tcl` fails. I'll have investigate further what might cause this.

```
    Start 194: test_XPlusYPlusZ.tcl
1/2 Test #194: test_XPlusYPlusZ.tcl .............***Failed    5.95 sec
-- Executing: tclsh test_XPlusYPlusZ.tcl 1 1 1
-- Output:

FAILED : Max difference in Pressure
    Minimum significant digits at (  2,   1,   2) =  0
    Correct value 3.401176e-01
    Computed value -3.168191e-01
    Difference 6.569367e-01
    Maximum absolute difference = 3.448811e+00
test_XPlusYPlusZ : FAILED

and

    Start 196: test_XYZTPlus1.tcl
2/2 Test #196: test_XYZTPlus1.tcl ...............***Failed    6.67 sec
-- Executing: tclsh test_XYZTPlus1.tcl 1 1 1
-- Output:

FAILED : Max difference in Pressure
        Minimum significant digits at ( 14,  14,   0) =  2
        Correct value 1.001358e+00
        Computed value 1.002561e+00
        Difference 1.202778e-03
        Maximum absolute difference = 1.239370e-03
FAILED : Max difference in concen
        Minimum significant digits at ( 14,  14,   0) =  2
        Correct value 1.001358e+00
        Computed value 1.002561e+00
        Difference 1.202778e-03
        Maximum absolute difference = 1.239370e-03
test_XYZTPlus1 : FAILED 
```

co-authored with: @skollet


smithsg84:
The 'ExactSolution' is implemented using DirichletBC so will be impacted by changing the calculation.   As will several other types.

https://github.com/parflow/parflow/blob/02c9da3bd3c4e62cdea9e768ce330c66aa3bef94/pfsimulator/parflow_lib/bc_pressure_package.c#L322

          SetupPatchInterval(ExactSolution,
          {
            NewBCPressureTypeStruct(ExactSolution, interval_data);

            BCPressureDataBCType(bc_pressure_data, i) = DirichletBC;

            GetTypeStruct(ExactSolution, data, public_xtra, i);

            ExactSolutionFunctionType(interval_data)
              = (data->function_type);

            BCPressureDataIntervalValue(bc_pressure_data, i, interval_number)
              = (void*)interval_data;
          }); 

The user supplied type (public_xtra->input_types[]) is mapped to an internal type to reduce the number of internal BC types when the implementations are basically the same; only the source of data is different.

kvrigor:
Thanks for the hints @smithsg84. Does this mean that we should simply update the reference `pfb` files for the failing tests? Doing so and running the tests on our cluster where we built ParFlow with Intel toolchain made all tests pass.

Oddly one test fails on GitHub (`test_XPlusYPlusZ.tcl`) despite seeing all tests pass on our cluster. Perhaps generating `pfb` files is sensitive to the environment which produces it? I will regenerate the reference results on an Ubuntu 22.04 environment and check whether all tests could still pass on our cluster as well as on GitHub.

smithsg84:
Yes we will need to update the regression files with this change since we would potentially expect changes on any test using Dirichlet boundary conditions.   I'm a bit surprised more of the tests don't see changes in the suite.

There is one test that fails with changes of compiler and Intel was causing issues for me.   I need to verify that this is the same test I see issues with.   I was not able to track down why one of our tests seems to exhibit this issue.  To be consistent the regression tests should be passing in the GitHub environment.

kvrigor:
Finally all GitHub tests could now pass after replacing the pressure file for `test_XPlusYPlusZ.tcl` with the one generated from ParFlow-Ubuntu 22.04. Unfortunately, these tests on our cluster fail: 

```
1/2 Test  #65: crater2D_vangtable_linear.tcl ....***Failed    1.55 sec
-- Executing: tclsh crater2D_vangtable_linear.tcl 1 1 1
-- Output:

FAILED : Max difference in ssat
        Minimum significant digits at (  0,   0,   0) =  0
        Correct value 1.000000e+00
        Computed value 0.000000e+00
        Difference 1.000000e+00
        Maximum absolute difference = 1.000000e+00
crater2D : FAILED

2/2 Test #193: test_XPlusYPlusZ.tcl .............***Failed    4.38 sec
-- Executing: tclsh test_XPlusYPlusZ.tcl 1 1 1
-- Output:

FAILED : Max difference in Pressure
        Minimum significant digits at (  5,   2,   6) =  0
        Correct value 1.469187e-02
        Computed value -1.492561e-02
        Difference 2.961748e-02
        Maximum absolute difference = 1.400887e+00
test_XPlusYPlusZ : FAILED
```

For `test_XPlusYPlusZ.tcl`, I noticed that it generates different pressure files depending on whether ParFlow is built with Intel or GCC. Interestingly, `pfclm_sc.tcl` doesn't have this problem - it produces bit-for-bit exact regression files on both toolchains.

[test_XPlusYPlusZ.out.press.00001.zip](https://github.com/parflow/parflow/files/11881548/test_XPlusYPlusZ.out.press.00001.zip)

```bash
# ParFlow built w/ GNU mpicc 11.3.0
$ sha256sum gcc_test_XPlusYPlusZ.out.press.00001.pfb 
27e9a40d7c3788fd439433aa42710e023067599043a4cdc20e8c9e6ff771c20e  gcc_test_XPlusYPlusZ.out.press.00001.pfb

# ParFlow built w/ Intel mpicc 2021.4.0 20210910
$ sha256sum intel_test_XPlusYPlusZ.out.press.00001.pfb
d272b3084853b316bf6776e9dd46373df0d0d48740b1abca08451c6b2bfdf9d2 
 intel_test_XPlusYPlusZ.out.press.00001.pfb
```

These inconsistent test results caused by different compilers may be masking a deeper problem which I could further investigate. Should we address this problem here or perhaps raise a separate GitHub issue for this?

smithsg84:
I think we should create new issue, this has been around for several years.

I looked into a bit but didn't find anything, checked for array bounds etc.   If you want to look into this that would be great.   

I think the crater2D_vangtable_linear is something I haven't seen.   Is it the entire file that is wrong or isolated cells?

smithsg84:
Started new issue to track:

https://github.com/parflow/parflow/issues/474

kvrigor:
> I think the crater2D_vangtable_linear is something I haven't seen. Is it the entire file that is wrong or isolated cells?

This baffles me as well. I'll set up a clean environment and see if I could reproduce this error.

skollet:
> Yes we will need to update the regression files with this change since we would potentially expect changes on any test using Dirichlet boundary conditions. I'm a bit surprised more of the tests don't see changes in the suite.
> 

The fix only concerns the bottom and top DirEquilRefPatch. Most test using DirEquilRefPatch apply left, right, front, bottom. Below is a list of tests in test/ that are impacted by the fix. In examples/ DirEquilRefPatch is not applied to bottom/top in any of the examples.

var_dz_1D, pfclm_sc, clm-reuse (z-lower commented out though), Dirichlet (new test case part of the fix)

smithsg84:
Thanks for tracking down the tests that use this boundary condition.

The setup in the test_XPlusYPlusZ example is not converging to a solution; not sure why that would lead to difference in output.   Both results should be the same non-converged result if the same calculation is being done.   Not a very good test case though.   I think the timestep needs to be smaller.   test_XYZTPlus1 is converging but has a 0.001 timestep vs 1.0 used in the other test_X* examples.

I don't think this is the root cause of the difference between the compiles.

kvrigor:
I did a clean rebuild of ParFlow and then re-ran the tests. The `crater2D_vangtable_linear.tcl` failure is now gone and the only failing test is `test_XPlusYPlusZ.tcl` (same error as before). 

```
194/214 Test #194: test_XPlusYPlusZ.tcl .............................***Failed    5.76 sec
-- Executing: tclsh test_XPlusYPlusZ.tcl 1 1 1
-- Output:

FAILED : Max difference in Pressure
	Minimum significant digits at (  5,   2,   6) =  0
	Correct value 1.469187e-02
	Computed value -1.492561e-02
	Difference 2.961748e-02
	Maximum absolute difference = 1.400887e+00
test_XPlusYPlusZ : FAILED

CMake Error at ${PARFLOW_SRC}/cmake/modules/RunParallelTest.cmake:33 (message):
  Test Failed: output indicated FAIL
Call Stack (most recent call first):
  ${PARFLOW_SRC}/cmake/modules/RunParallelTest.cmake:87 (pf_exec_check)
```

Would this PR be safe to merge despite having failed tests under a different compiler? Or should we address #474 first before merging?

[icc2021.6.0_test_results.txt](https://github.com/parflow/parflow/files/12327545/icc2021.6.0_test_results.txt)
[icc2021.4.0_test_results.txt](https://github.com/parflow/parflow/files/12327546/icc2021.4.0_test_results.txt)

kvrigor:
Out of curiosity I ran the tests on `parflow-master`, and there `test_XPlusYPlusZ.tcl` also fails. So #474 seems to have been undetected for quite some time until it was revealed in this PR.

smithsg84:
Yes issue #474 is unrelated and predates this issue.

reedmaxwell:
Reviewed and approved.  Thanks @kvrigor @skollet and @smithsg84 for fixing this!

reedmaxwell (approved):

PR #475, opened 2023-07-06
Speed up vegm file reading

arbennett:
This is a relatively small change, which just speeds up reading large vegm files in the python functions. I'm not really sure how to implement a test for this but have verified that it at least produces bit-for-bit reproducibility with the old function as shown by my script below:

```python
def read_vegm_new(file_name):
    # Assume first two lines are comments and use generic column names
    df = pd.read_csv(file_name, delim_whitespace=True, skiprows=2, header=None)
    df.columns = [f'c{i}' for i in range(df.shape[1])]

    # Number of columns and rows determined by last line of file
    nx = int(df.iloc[-1]['c0'])
    ny = int(df.iloc[-1]['c1'])
    # Don't use 'x' and 'y' columns
    feature_cols = df.columns[2:]
    # Stack everything into (ny, nx, n_features)
    vegm_array = np.stack(
        [df[c].values.reshape((ny, nx)) for c in feature_cols], axis=-1
    )
    return vegm_array

def read_vegm_old(file_name):
    with open(file_name, 'r') as rf:
        lines = rf.readlines()
    
    last_line_split = lines[-1].split()
    x_dim = int(last_line_split[0])
    y_dim = int(last_line_split[1])
    z_dim = len(last_line_split) - 2
    
    # To be consistent with ParFlow-python xy indexing, x should represent columns and y rows
    vegm_array = np.zeros((y_dim, x_dim, z_dim))
    # Assume first two lines are comments
    for line in lines[2:]:
        elements = line.split()
        x = int(elements[0])
        y = int(elements[1])
        for i in range(z_dim):
            vegm_array[y - 1, x - 1, i] = elements[i + 2]
    return vegm_array

file_name = '/hydrodata/PFCLM/CONUS1_baseline/inputs/drv_vegm.fixed.dat' 
```

```
%%time 
old = read_vegm_old(file_name)
```
CPU times: user 45.1 s, sys: 879 ms, total: 46 s
Wall time: 46.1 s

```
%%time 
new = read_vegm_new(file_name)
```
CPU times: user 7.69 s, sys: 906 ms, total: 8.59 s
Wall time: 8.62 s

```
np.sum(old - new)
```
0.0


smithsg84 (approved):
Thanks for the performance improvements.

PR #476, opened 2023-07-10
Python tests optional dependencies fixes #367

### LINKED ISSUE #367: Python regression tests and optional dependencies
The Python regression tests are not checking if optional packages (hypre, silo etc) are installed when running the regression tests.   Looks like all tests are always being run which results in failures when optional packages are not present.

gartavanis:
Added guards to check if optional packages are installed when running python regression tests

The following tests should not run when HYPRE is missing: 
py_prefix_naming
py_serial_runs
py_simple-mask
py_washita
py_washita_clm_keys
py_clm_reuse
py_clm, py_clm_varDZ
py_clm.jac
py_clm_forc_veg
py_clm_4levels
py_clm_slope

The following tests should not run when CLM is missing: 
py_washita
py_washita_clm_keys 

The following tests should not run when SILO is missing: 
py_default_overland.pfmg_octree.jac
py_clm
py_clm_varDZ
py_clm.jac
py_clm_forc_veg
py_clm_4levels

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -10,6 +10,7 @@ ParFlow development and bug-fixes would not be possible without contributions of
 * Documentation updates
 * 
 
+
 ## User Visible Changes
 
 ### DockerHub Container has Python and TCL support.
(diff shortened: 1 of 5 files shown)


smithsg84 (approved):
Thanks for fixes.

PR #477, opened 2023-07-20
Python regression test checks

gartavanis:
Added module utils.py in test/python. It contains the python equivalents of the tcl procedures pftestFile, pftestFileWithAbs, MSigDiff.

Completed the regression checks in defaults_richards_wells.py.

### DIFF
--- a/test/python/default_richards_wells.py
+++ b/test/python/default_richards_wells.py
@@ -6,11 +6,11 @@
 
 from parflow import Run
 from parflow.tools.fs import mkdir, get_absolute_path
+from utils import py_pftestFile
 
-drich = Run("default_richards_wells", __file__)
-
+run_name = "default_richards_wells"
+drich = Run(run_name, __file__)
 #---------------------------------------------------------
-
 drich.FileVersion = 4
 
 drich.Process.Topology.P = 1
@@ -348,7 +348,33 @@ drich.Solver.Linear.Preconditioner = 'MGSemi'
 #-----------------------------------------------------------------------------
 # Run and Unload the ParFlow output files
 #-----------------------------------------------------------------------------
+new_output_dir_name = get_absolute_path('test_output/drich_w')
+correct_output_dir_name = get_absolute_path('../correct_output')
+mkdir(new_output_dir_name)
+drich.run(working_directory=new_output_dir_name)
 
-dir_name = get_absolute_path('test_output/drich_w')
-mkdir(dir_name)
-drich.run(working_directory=dir_name)
+passed = True
+
+filename = f"/{run_name}.out.perm_x.pfb"
+if not py_pftestFile(new_output_dir_name + filename, correct_output_dir_name + filename, "Max difference in perm_x"):
+    passed = False
+filename = f"/{run_name}.out.perm_y.pfb"
+if not py_pftestFile(new_output_dir_name + filename, correct_output_dir_name + filename, "Max difference in perm_y"):
+    passed = False
+filename = f"/{run_name}.out.perm_z.pfb"
+if not py_pftestFile(new_output_dir_name + filename, correct_output_dir_name + filename, "Max difference in perm_z"):
+    passed = False
+
+timesteps = ["00000", "00001", "00002", "00003", "00004", "00005"]
+for i in timesteps:
+    filename = f"/{run_name}.out.press.{i}.pfb"
+    if not py_pftestFile(new_output_dir_name + filename, correct_output_dir_name + filename, f"Max difference in Pressure for timestep {i}"):
+        passed = False
+    filename = f"/{run_name}.out.satur.{i}.pfb"
+    if not py_pftestFile(new_output_dir_name + filename, correct_output_dir_name + filename, f"Max difference in Saturation for timestep {i}"):
+        passed = False
+
+if passed:
+    print(f"{run_name} : PASSED")
+else:
+    print(f"{run_name} : FAILED")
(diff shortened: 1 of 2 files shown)


smithsg84:
I believe this is missing a backend that can load the pfb files?

Locally I get a failure:

Error: Failed to load data from files. did not find a match in any of xarray's currently installed IO backends ['netcdf4', 'h5netcdf', 'scipy', 'pydap', 'rasterio', 'zarr']. Consider explicitly selecting one of the installed engines via the ``engine`` parameter, or installing additional IO dependencies, see:
https://docs.xarray.dev/en/stable/getting-started-guide/installing.html
https://docs.xarray.dev/en/stable/user-guide/io.html
Traceback (most recent call last):
  File "/home/smith84/projects/parflow/failed-build/pf-build/parflow/test/python/default_richards_wells.py", line 359, in <module>
    if not py_pftestFile(new_output_dir_name + filename, correct_output_dir_name + filename, "Max difference in perm_x"):
  File "/home/smith84/projects/parflow/failed-build/pf-build/parflow/test/python/utils.py", line 76, in py_pftestFile
    if len(result) != 0:
TypeError: object of type 'NoneType' has no len()
<end of output>

smithsg84:
There is a read_pfb method in the library, why don't we use that?

smithsg84:
OK there is pf backend for xarray here:

parflow/tools/pf_backend.py

this doesn't appear to be hooked in correctly to xarray so the open can use it.

gartavanis:
>>> We may need to start explicitly listing all the versions in the requirements.txt file rather than just getting the latest.

      I agree. We had a similar problem here on verde about a month ago when some features in numpy got deprecated. Because pftools does not list version requirements, numpy was silently updated when we installed another package, which broke pftools.
________________________________

test/python/utils.py
```diff
@@ -0,0 +1,121 @@
+import os
```
smithsg84, line 1: I think the testing methods should be in the parflow package.

maybe

parflow.tools.compare

Putting the TCL utils in the test directory is a bit ugly when we started to add multiple test subdirectories.    People started to copy the file :(

test/python/default_richards_wells.py
```diff
@@ -348,7 +348,33 @@
 #-----------------------------------------------------------------------------
 # Run and Unload the ParFlow output files
 #-----------------------------------------------------------------------------
-
-dir_name = get_absolute_path('test_output/drich_w')
-mkdir(dir_name)
-drich.run(working_directory=dir_name)
+new_output_dir_name = get_absolute_path('test_output/' + run_name)
```
smithsg84, line 351: I made a change here to make the output directory the run_name.   originally it was 'drich' which doesn't make as much sense to me.   That's the name of the temp Run object.

Does that make sense to you?   I'm not sure why the Python output was named that way originally.

test/python/default_richards_wells.py
```diff
@@ -6,11 +6,11 @@
 
 from parflow import Run
 from parflow.tools.fs import mkdir, get_absolute_path
+from utils import py_pftestFile
 
-drich = Run("default_richards_wells", __file__)
-
+run_name = "default_richards_wells"
```
smithsg84, line 11: Is this easier?   Saves having to change the runname all the time?  Not super sure, might be less clear.

import pathlib
run_name = pathlib.Path(__file__).stem

pftools/python/requirements.txt
```diff
@@ -1,5 +1,8 @@
 PyYAML==6.0.1
+ecmwflibs
```
smithsg84, line 2: I'm going to remove my additions here since you changed the reader.   I don't think we are using these readers.

test/python/utils.py
```diff
@@ -0,0 +1,121 @@
...
+from parflow.tools.io import read_pfb
+
+def py_MSigDiff(file1, file2, m, abs_zero=0.0):
```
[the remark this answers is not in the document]
gartavanis: Speed is not an issue with read_pfb, only with xarray.open_dataarray.

The file is not opened and read 5 times. But each test loads and reads multiple pfb files (e.g. defaults_richards_wells loads 12 different pfbs). 

In the TCL version, the caller opens the files and passes data arrays to MSigDiff. But the number of times files are opened and read is the same in both TCL and python.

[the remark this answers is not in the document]
gartavanis: I'll change that to lowercase instead. 

I added the py_ prefix to distinguish these functions from their equivalent tcl versions. But I can remove it - hopefully it should be clear which one is called.

test/python/utils.py
gartavanis: I agree - I'll move it. I was trying to copy the tcl structure but it fits better out of the test directory.

test/python/default_richards_wells.py
gartavanis: agreed - drich seemed random.

test/python/default_richards_wells.py
gartavanis: The test files are already written, so I'm not sure if that saves time. Writing the run_name explicitly might be clearer too.

test/python/utils.py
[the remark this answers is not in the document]
smithsg84: Isn't the file read in py_MSigDIff with read_pfb and in each of the xr.open_dataarray() calls below in the py_pftestFile methods.

I think we should be doing a single read_pfb on the files and then just accessing the returned array.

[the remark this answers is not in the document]
gartavanis: you are correct - I thought I replaced all xr.open_dataarray calls already.

smithsg84 (approved):

PR #481, opened 2023-07-28
Add Python style guidelines

### LINKED ISSUE #480: Python Style Guidelines
We do not have any documentation on our Python style guidelines.   Add some info to the contributing documentation.

smithsg84:
Closes #480


alanlewis85:
Unless I’m mistaken, the LGPL and GPL licenses are only compatible when the combined code is GPL. That is, if ParFlow is LGPL and accepts GPL code, then ParFlow becomes GPL.

alanquits:
I was referring to this statement, "It is strongly preferred to use GNU General Public License version 2.1 for
contributions since other licenses will have to be checked to see if including them is allowed."

LGPL and GPL are compatible in the sense that they can be used in the same project. However, the resulting combination is licensed under GPL. So, if ParFlow accepts GPL code, ParFlow is technically under the GPL license (not LGPL). It's a minor detail that likely won't really affect ParFlow, but it does have implications for potential use in closed-source, commercial applications.

alanquits:
Thanks for the clarification. “Compatible” Is used both ways.

My comment was more that the text says that ParFlow is released under LGPL, and then appears to request contributions under the GPL.

Issue #482, opened 2023-08-02
BC defined with PFB file generates error | malloc_consolidate(): unaligned fastbin chunk detected


jsacerot:
I am trying to run ParFlow coupled with CLM and I am getting the standard error file full of the following lines (repeated for each task per node):

```
corrupted size vs. prev_size in fastbins
malloc_consolidate(): unaligned fastbin chunk detected
malloc_consolidate(): unaligned fastbin chunk detected
[centaur.cluster:mpi_rank_35][error_sighandler] Caught error: Aborted (signal 6)
```

@smithsg84  What could be the issue? It seems an improper memory handling.

All the output files were generated, so the problem occurred at the very end of the ```pfrun``` execution when ParFlow was wrapping the simulations (```pfundist``` was not executed). I also noticed that CLM outputs were not created in the folder set using the  ```Solver.CLM.CLMFileDir``` key. The folder was not even created.

Here is the full log file as a reference and my tcl script.
[Results.zip](https://github.com/parflow/parflow/files/12245003/Results.zip)

Thanks!

jsacerot:
@smithsg84 
**Update:**
The issue appears when using a pfb file to define the left patch boundary condition by passing ```PressureFile``` to the ```BCPressure.Type``` key. CLM has nothing to do with this issue.

jsacerot:
@smithsg84

**Update [09/06/2023]:**
I recompiled ParFlow 3.11 (the version that worked well before the cluster upgrades) using the new environment and got the same errors. So, the issue is likely related to the compilers/MPI versions that ParFlow does not like.

jsacerot:
@smithsg84 @reedmaxwell 

**Update [09/08/2023]:**
I ran the model at the UCR HPC and got the same errors (see attached file). Any ideas on how to fix this? The issue only appears when using a PFB file to define a constant pressure BC. It occurs after the ParFlow simulation ends, which may be during the file undistribution.

[slurm.245326.err.zip](https://github.com/parflow/parflow/files/12562776/slurm.245326.err.zip)

smithsg84:
I have been able to recreate this bug locally.   Working on fix.

jsacerot:
> I have been able to recreate this bug locally. Working on fix.

Thank you, @smithsg84.

PR #486, opened 2023-09-12
Memory corruption when using the PressureFile option

smithsg84:
The PressureFile option (and others) were causing memory corruption due to multiple frees of the filenames.   Removed the incorrect free calls in the shutdown logic.

Removed macros that wrap code since this makes the code harder to maintain and achieved nothing.


smithsg84:
Fixes #482

PR #488, opened 2023-09-13
Draft: Reformat

smithsg84:
Reformat code with latest uncrustify; formatting was getting very poor and inconsistent.


smithsg84:
This was completed in #521

Issue #490, opened 2023-09-16
ParFlow cannot read EvapTrans files with more than 5 digits in their names


jsacerot:
I am having issues using EvapTrans files in transient mode:

```
pfset Solver.EvapTransFileTransient                                       True
pfset Solver.EvapTrans.FileName                  $env(EVAPTRANS_DIR)/EvapTrans
```
This is the error I am getting:

```
Aborting in file </home/jsat/ParFlow-3.12/parflow/pfsimulator/amps/common/amps_ffopen.c> line : 205
AMPS Error
```

After performing a couple of tests changing the number of digits in the EvapTrans file names, it seems that **ParFlow only reads files with up to 5 digits on their names (e.g., EvapTrans.06554.pfb), limiting the number of files that can be read to 99,999, which, if correspond to hourly data, are only 11.42 years (assuming 365 days per year)**. This is probably a bug because other input files are allowed to have 6 digits in their names.

jsacerot:
I noticed that ParFlow tries to read an EvapTrans file for t = 0 (initialization), which should not be the case unless there is a specific reason.

jsacerot:
A key should be added in the ParFlow input keys to enable users to define the number of digits in file names for both input and output files. Allowing some files to have up to five and some with up to six is impractical.

PR #491, opened 2023-09-16
Update solver_richards.c to correct digit number in EvapTrans file names

jsacerot:
Fix minor bug limiting EvapTrans file names to five digits instead of six as forcing input files. See [issue 490](https://github.com/parflow/parflow/issues/490).


smithsg84:
This change will break some workflows that depend on the existing 5 digits in the file names.   The original PF files are using 5 digits some newer CLM files are using 6 digits.

Will keep the issue open for a more complete fix but closing this MR so we can have a more complete change.

PR #501, opened 2024-01-02
Hotfix/var dz wells

westb2:


smithsg84:
The PR doesn't compile, converting to draft.

pfsimulator/parflow_lib/reservoir_package.c
westb2, line 1: The changes in this file are for two reasons. Firstly, I wanted reservoirs and wells to use the same functions behind the scenes as the functions here were only meant as a temporary placeholder until there was something more permanent available. Secondly, during testing I found a directionality bug with reservoirs and overlandkinematic which has now been fixed.

pftools/python/parflow/tools/tests/pfb_summary.py
westb2, line 1: I think my editor did some auto formatting of files I opened that accidentally got committed. If wanted, I can manually roll this back.

test/tcl/default_single.tcl
```diff
@@ -10,6 +10,7 @@ namespace import Parflow::*
 # File input version number
 #-----------------------------------------------------------------------------
 pfset FileVersion 4
+pfwritedb default_single
```
smithsg84: This not write a valid database file, why is this here?

pfsimulator/parflow_lib/well_package.c
```diff
@@ -510,17 +569,30 @@ void         WellPackage(
...
+              {
+                subgrid_volume = CalculateSubgridVolume(new_subgrid, problem_data);
+              }
+            }
+            else
+            {
+              subgrid_volume = nx * ny * nz * dx * dy * dz;
+            }
+#ifdef PARFLOW_HAVE_MPI
+            // here I am making the assumption that indices need to be positive so we can use amps_max
```
smithsg84: What does the indices have to do with this reduction?   volumes > 0 so find use max?

test/correct_output/multi_column_var_dz_with_well.out.press.00010.pfb.dist
```diff
@@ -0,0 +1 @@
+0
```
smithsg84: Dist files are not normally committed, are these dist files need for some reason?

smithsg84 (changes requested):
Hi @westb2.

Sorry about losing track of this PR.   I'm confused by the MPI reduces going on here.

Steve

test/python/var_dz_with_well.py
```diff
@@ -0,0 +1,508 @@
...
+
+import parflow as pf
+
+vardz = pf.Run("model", __file__)
+
+# ---------------------------------------------------------
+
+vardz.FileVersion = 4
+
+vardz.Process.Topology.P = 1
```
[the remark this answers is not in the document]
westb2: If this file is testing too many things I would be happy to split it into a single column and multi column test, or have it be just a multicolumn if we don't think we need to test the single column test case. I had it in one place during development because it made it easier to be sure I was changing one thing at a time with each case

pfsimulator/parflow_lib/well_package.c
westb2: I believe that was my initial concern with this comment but after reviewing the code I don't believe this assumption is required anymore so will remove the comment.

pfsimulator/parflow_lib/well_package.c
```diff
@@ -202,18 +224,40 @@ void         WellPackage(
           rz = 0;
 
           process = amps_Rank(amps_CommWorld);
+#ifdef PARFLOW_HAVE_MPI
+//          // here I am making the assumption that indices need to be positive so we can use amps_max
```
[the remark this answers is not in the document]
westb2: My memory is that when we discussed a potential algorithm for this we were worried about the case of the well being split across multiple ranks as a particularly difficult edge case we were happy not supporting given it could only happen if people set R>1, and not with P or Q. That said, I believe my final implementation may have needed to handle this edge case anyways because the code leaves open the possibility of multiple subgrids on a single rank. I will try changing this reduction to a sum and adding a test case for an R split and see if it works.

westb2:
Hi @smithsg84 I believe I have addressed your requested changes in the PR but there are a couple things I might imagine that you want changed and would love to have your feedback at a high level before asking you to do an in depth review again. t:

1) The test case I wrote runs the same Parflow model multiple times with configuration changes and checks that the outputs match when appropriate, and that they don't match when they shouldn't. Namely, I make use of the fact that there is a both a ComputationalGrid.DZ and the dzScale list that we can manipulate to generate test cases that should be equivalent. Some of the test cases:

A single column setup should match another single column setup

A 2x2 column setup should match another 2x2 column setup. It should also continue matching if we distribute it across more ranks

A 2x2 column setup with the correction turned on should NOT match the same setup with the correction turned off.

To check each of these I have one file and I do a run, then change only the relevant keys and do another run and compare the two outputs. This results in many less test cases being committed as files, but potentially at the cost of legibility. 

Would you prefer each of these test cases is it's own file that checks against an expected output that is committed?

2) The looping logic is definitely a little hard to grok, I think in part because it has to interact with so many of parflow's macros. However, some of the functions I think could certainly be reused in other parts of the code (wells and reservoirs may not be the only thing in the long run that need a vardz correction). Are they in the right part of the codebase and is there something I could do to improve long term discoverability and/or legibility?

Thanks in advance for your time

test/python/default_single.py
```diff
@@ -10,6 +10,7 @@
 run_name = "default_single"
 default_single = Run(run_name, __file__)
 
+
```
westb2: What is our philosophy on whitespace changes like this getting shipped? Asking because they show up in a couple files, presumably because I was interacting with them in some temporary way. Happy to go through and manually roll them back if we are bothered but figured I would at least ask what the standard is first

smithsg84:
@westb2 

The single column subtests are not behaving as I would expect.   If I hack code to make the well volume 8.0, adding a 

```
subgrid_volume = 8.0;
```
After the reduction the single column tests still pass the pressure comparison test.     Shouldn't that cause a failure?

Checking the well size/volume in the output shows volume is 8.0 with the hack.

```
grep size test_output/*/*.out.txt
test_output/var_dz_with_well_1/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_2/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_3/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_4/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_5/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_5/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_6/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_6/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_7/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_7/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_8/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_8/var_dz_with_well.out.txt:Node 0:    size = 8.000000
test_output/var_dz_with_well_9/var_dz_with_well.out.txt:Node 0:    size = 10.000000
test_output/var_dz_with_well_9/var_dz_with_well.out.txt:Node 0:    size = 10.000000
```

The subtests are only failing for the multi-column pressure check, shouldn't single column also fail?

westb2:
Hi @smithsg84 ,

Took a little more finicking on my end, but all the test cases are passing now, including the ones distributed in Z. And when hardcoding an improper value (e.g. 8.0) all the tests fail properly on my machine, including the single column ones.

I also made the change of generating the correct output files by hard coding the expected volume calculation (11.0) so we know not just that the test cases are internally consistent across different equivalent permutations of the var-dz inputs, but that they are correct (which I was previously eyeballing in a manual test). Don't know why I didn't think of this earlier. 

I realized when looking at the errors you had caught part of what had happened was I had tested some stuff manually, but then made some errors consolidating that into one test case. Because some changes have been made since I last ran these I took the time to reverify them today. Specifically, I checked that the that the values for index_upper and index_lower respond correctly to changes in the var_dz inputs. Because these values are never exposed back up to python it's a bit hard to figure out how to work them into the unit tests, but I think the tests we have now are enough.

Thanks for the time you have spent to look this over and in constructing new tests. For the distribution in z in particular, I remember that we thought our proposed algorithm would not work and we should leave it out of scope, but it seems that between the changes made since design and catching the Amps_Max error that case is handled now, which is nice. If I test switching back to amps_max, those are the only tests that break so I think that's why I didn't catch that.

Sorry for letting this drag out so long - I have done a pretty bad job of managing the communication via github and getting back on this promptly once you look at it. I will do better at that in the future.

Lastly, I'm planning on talking with Laura about the concern I raised in the dev meeting (just so you know that issue is being surfaced). I imagine she will have a better idea than me if there actually is cause for concern.

smithsg84 (approved):
Thanks for patching and fixing

PR #526, opened 2024-06-03
Remove local cub dependency/ directory to utilize CUDA-Provided version

mfahdaz:
This PR removes the dependency on the local cub directory. Since CUDA 11.0, CUB is [officially supported and included](https://github.com/NVIDIA/cub/releases/tag/1.9.8) in the CUDA Toolkit. Therefore, there's no need to keep a local version of CUB, as it's causing conflicts with CUDA builds.

- Motivation and Context

The purpose of this change is to avoid conflicts with the official CUB library bundled with CUDA, particularly from CUDA 12.0. Removing the local version of CUB aligns with the standard CUDA setup and streamlines the build process, ensuring better compatibility and reducing maintenance overhead.

- Changes Made

  - Removed the `cub` directory from the repository.
  - Updated the relevant build scripts and CMake files to remove the references to local version of CUB.
  - Adjusted all instances where CUB is included or referenced to use the version provided by CUDA.


smithsg84 (approved):
Thanks for update

PR #531, opened 2024-07-02
Mg semi of jacobian precon

smithsg84:


test/python/overland_tiltedV_MGSemiTest.py
```diff
@@ -0,0 +1,511 @@
...
+#overland.Solver.Linear.Preconditioner.PCMatrixType = 'PFSymmetric'
+
+
+run_name = "TiltedV_OverlandKin_JacFalse_MGSemi"
+overland.set_name(run_name)
+print("##########")
+print(f"Running {run_name}")
+new_output_dir_name = get_absolute_path('test_output/' + f"{run_name}")
+mkdir(new_output_dir_name)
+overland.run(working_directory=new_output_dir_name)
```
smithsg84, line 357: Need to check output of the run.   Simply running doesn't indicate very much.   No idea what the result converged to.

reedmaxwell:
The MGSemi overland flow branch is ready to go and can be merged after review.  @gartavanis  and I worked through some GPU/CUDA tests and I fixed a bug in one of the other tests that was causing it to fail. 

Here’s a quick summary:

- `richards_jacobian_eval.c` has been updated to run with the true jacobian and MGSemi.  This allows efficient overland flow simulation with that preconditioner and now provides full PF capabilities on GPUs namely OverlandFlow, OverlandKinematic, and OverlandDiffusive.  This is backwardly compatible and preserves all prior configurations (jacobian false, PFMG and other Hypre preconditions, etc).

-All the standard PF tests appear to be passing on my laptop, except for the ones w/ processor configurations requiring more cores than I have

-Two new tests, overland_tiltedV_MGSemiTest and LW_Test_MGSemi have been added that check all new functionality; these tests also work on CPU and GPU (using the CUDA backend). Additional testing was conducted with a CONUS2 HUC using subsettools and MGSemi, this test runs on CPU and GPU

test/python/overland_tiltedV_MGSemiTest.py
```diff
@@ -402,7 +402,16 @@ def CheckOutput(run_name, correct_root, correct_output_dir_name):
...
+## need to change the channel slopes for Overland Flow which is cell-centered
+overland.TopoSlopesX.Type = 'Constant'
+overland.TopoSlopesX.GeomNames = 'left right channel'
+overland.TopoSlopesX.Geom.left.Value = -0.01
+overland.TopoSlopesX.Geom.right.Value = 0.01
+overland.TopoSlopesX.Geom.channel.Value = 0.00
+overland.TopoSlopesY.Type = 'Constant'
+overland.TopoSlopesY.GeomNames = 'domain'
+overland.TopoSlopesY.Geom.domain.Value = 0.01
+overland.Patch.z_upper.BCPressure.Type = 'OverlandFlow'
```
reedmaxwell, line 414: this code was added because modified test was not actually running OverlandFlow, it was running OverlandKinematic, however it was comparing to the correct output from original test (which was correclty running with OverlandFlow) and that is why it was failing.

gartavanis:
I did a bit of cleaning up, removed .dist files that were not needed from correct_output, fixed some build errors and removed unused variables that were causing compiler warnings. I think this is ready to be reviewed and merged.

PR #537, opened 2024-07-31
Remove very slow CUDA tests

gartavanis:
This PR removes python tests with processor topology "3 3 3" when run on the GPU, as they take a really long time to complete. The equivalent tcl tests with the same topology were already removed a long time ago (the code addition is the same as in test/tcl/CMakeLists.txt)


test/python/CMakeLists.txt
```diff
@@ -139,6 +139,10 @@ foreach(inputfile ${PARALLEL_3DTOPO_TESTS})
   pf_add_py_parallel_test(${inputfile} 1 4 1)
   pf_add_py_parallel_test(${inputfile} 4 1 1)
   pf_add_py_parallel_test(${inputfile} 2 2 2)
+  if(((${PARFLOW_HAVE_CUDA}) OR (${PARFLOW_HAVE_KOKKOS}) OR (${PARFLOW_HAVE_OMP})))
+    #These tests should pass with CUDA or OMP, but they take a very long time
+    continue()
```
smithsg84, line 144: Rather than do a continue() which is a bit counter intuitive, is there a reason not to invert the if and put the pf_add in the if....

```
if(not( ((${PARFLOW_HAVE_CUDA}) OR (${PARFLOW_HAVE_KOKKOS}) OR (${PARFLOW_HAVE_OMP})) ))
    pf_add_py_parallel_test(${inputfile} 3 3 3)
endif
```
breaking out of loops should be used for special cases.

smithsg84 (changes requested):
Question about how logic is being changed.

smithsg84 (approved):

PR #544, opened 2024-08-15
Fix CMakelists for tests requiring HYPRE

gartavanis:


smithsg84 (approved):
Thanks for fixing hypre tests.

PR #547, opened 2024-09-19
Feature/Groundwater Flow BC

tikoh-station:
Adds new feature: GroundwaterFlow BC.

This implementation is based on the paper [M. Rahman et. al., 2018, "Towards a computationally efficient free-surface groundwater flow boundary condition for large-scale hydrological modelling"].


tikoh-station:
**State of the Pull-Request:**
The GroundwaterFlow BC works and produces the desired results for the infiltration experiment in [Rahman.2018].
Two parameters must be defined to use the GroundwaterFlow BC:
- Specific Yield 
- Aquifer Depth

Currently, the parameters are set to constant values through the TCL script.

**TODO List:**
- Some code polishing
- Add Heterogeneous Groundwaterflow parameter input
- Currently, this feature only works for domains with a flat bottom. To fix this, surface-surface interactions `JC` matrix of the jacobian in `richards_jacobian_eval.c` file must be modified to include the bottom surface as well.
- Fix and finish documentation files
- Add test case to the list of test cases

docs/manuals/files.tex
```diff
@@ -2510,7 +2510,7 @@ \subsection{Boundary Conditions: Pressure}
 change according to any cycle.  Instead, time dependence is handled by
 evaluating at the time the boundary condition value is desired.
 The solution is specified by using a predefined function (choices are described
-below).  NOTE: These last six types of boundary condition input is for
+below).  The choice {\bf GroundwaterFlow} turns on groundwater interactions with a deep aquifer as described in \cite{Rahman2018}. The aquifer is set with the Specific Yield and Thickness parameters, detailed below.   NOTE: These last seven types of boundary condition inputs are for
```
smithsg84, line 2513: There are some limitations noted in a comment on github about only supporting flat bottom.   At a minimum those should be documented.

pfsimulator/parflow_lib/compute_bottom.c
```diff
@@ -0,0 +1,111 @@
...
+ * Only works with 1 subgrid per task.
+ *
+ * This assumes number of processors is 1 in Z; assumes
+ * that the entire Z column is on a single task.
+ *
+ */
+
+#include "parflow.h"
+
+void ComputeBottom(
```
smithsg84, line 45: Computing top and bottom can be merged into a single function; this avoids doing 2 loops.

pfsimulator/parflow_lib/parameter_union.h
```diff
@@ -0,0 +1,233 @@
...
+*  Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307
+*  USA
+**********************************************************************EHEADER*/
+
+/*
+ * ParameterUnion looks for a parameter that might go by
+ * different aliases or assume different types of values.
+ *
+ * It can be used in cases where:
+ * - a physical variable may come under different names
```
smithsg84, line 34: What is the use case for aliasing?   Where are we doing this?

This will make input correctness much more complicated; does the python error checking support this?

pfsimulator/parflow_lib/parameter_union.h
```diff
@@ -0,0 +1,233 @@
...
+ * - different physical quantities can provide the same
```
smithsg84, line 35: Is there an example of this of an input database being used like this?

pfsimulator/parflow_lib/parameter_union.h
```diff
@@ -0,0 +1,233 @@
...
+ *  information, so when one is defined, all the others
+ *  are readily obtained
+ * - a variable may be defined through different inputs
+ *  (e.g. through a constant value or a filename).
+ *
+ * ParameterUnion will produce an error if none of the
+ * required keys is found. At least one key/alias must be
+ * defined in the input file. In case multiple aliases are
+ * defined in the input file, the first key to be found is
```
smithsg84, line 44: If we allow aliasing it should be an error to provide multiple aliases for the same value.   Highly likely the user doesn't intend this.

test/tcl/gfb_infiltration.tcl
```diff
@@ -0,0 +1,327 @@
+#  This runs the infiltration experiment using the GroundwaterFlowBC.
```
smithsg84, line 1: Need a Python test added to match this TCL test.

smithsg84 (changes requested):
Hi, 

Thanks PR request.   There are few high level things I made comments about.  Will do a more detailed review later.

Steve

pfsimulator/parflow_lib/parameter_union.h
smithsg84: Simplifying the coding for the Constant/PFBFile etc cases is good.

Part of the reason for explicitly setting the type explicit is this allows using the same attribute names for different types.   If new classes are added that also have same attributes then the attribute names alone wouldn't be sufficient unless they are named differently.   E.G. in the future if another type is added that also takes Value or FileName attributes.    Using simple attribute names is no longer future proof and I think this would require the attribute names be Type specific to disambiguate.

I'm not fully understanding how the union simplifies case of DeepAquiferBC being set from permeability, wouldn't you need a key to single that, like a Type key?   Key.Type = Permeability? 

I need to look over the python error checking, they wrote a little domain specific language to define the keys and I'm not sure it would handle aliasing.   I think they might use the Type attribute as well to check.

smithsg84: The Python error checking does not appear to be looking at the type and then checking to make sure that required attributes for that type are present.   It probably should do that.

smithsg84: Just thinking out loud....another item discussed at the start of the Python was enabling auto-complete in an editor for helping to build input files.   I wonder if having a Type would make that process flow better.   User auto-completes the Type and then auto-complete could know which attributes are required.

tikoh-station: I see. I understand the reasoning now. 

As for the permeability, my reasoning was that Parflow could be flexible enough not to expect any specific combination of input parameters, and display different behaviors depending on the information that it was provided. Permeability would be an additional optional key to provide futher customization of inputs. My idea was to generalize the union even further to allow a parameter defined from a combination of others. I am starting to realize that while this reduces the size of the Tcl and python scripts, it might not necessarily improve user experience.

What do you think is the best approach moving forward, having into consideration the future intended developements for the `DeepAquiferBC`?

tikoh-station: Can you elaborate on your comment about the Python error checking? Is this something to do with the `YAML` generated files for the python format?

tikoh-station:
Issue #588 `CLM Code Duplication` has been addressed.

PR #548, opened 2024-09-24
fixed tolerance issue on reservoir test case validation

westb2:
During work I noticed that at some point the correct pressure file for my reservoir test case with the OverlandFlow setting had been swapped for a different, incorrect pressure file somehow. More perniciously, the test case was still failing. I believe this to be because at some point during code review I switched from comparing the pressures using np.allclose() to a function we developed to do this that takes a custom tolerance. This tolerance was set too loosely. For the time being I have fixed the correct_output pressure file and reverted to checking with np.allclose() instead of fiddling with the tolerance to figure out what is appropriate.

When I discovered this I noticed we check some of our other test cases with the same tolerance settings e.g. example_single.py, default_single.py, and about 5 more. I believe this is where I got my setting from. We also check many other test cases with a tolerance that is not much more strict. We should probably tighten the tolerance on these test cases given that it gave my test a passing result with a pretty significantly different pressure field. If people agree, I'm happy to open an issue so we have it tracked.

I have validated that the current solve of the reservoir test case in question is happening correctly.


test/python/reservoir_mpi_test.py
```diff
@@ -337,13 +337,12 @@
 i=10   
 timestep = str(i).rjust(5, '0')
 
-sig_digits = 4
-abs_value = 1e-12
-
```
westb2, line 342: These are the tolerance settings in question. I believe the main issue is using too few significant digits

test/python/reservoir_mpi_test.py
```diff
@@ -337,13 +337,12 @@
...
 test_files = ["press"]
 for test_file in test_files:
     filename = f"/{run_name}.out.{test_file}.{timestep}.pfb"
-    if not pf_test_file_with_abs(test1_dir + filename, base_case_dir + filename, f"Max difference in {filename}", sig_digits, abs_value):
```
westb2, line 346: And this was the function I was using these tolerance settings in

smithsg84:
We intentionally moved away from tests like all_close due to issues with small numbers, like the comments in the docs.   It wasn't identifying sig differences in results.

https://numpy.org/doc/stable/reference/generated/numpy.allclose.html

The goal was to compare to a number of sig digits to avoid the scaling issue.   The abs_value was added to avoid numbers that are effectively zero.

If a tighter tolerance is desired why not increase the sig_digits?

smithsg84:
The actual bug here seems to be the Python check not finding differences.   The TCL version shows a difference:

FAILED : Max difference in Pressure for timestep 10
	Minimum significant digits at ( 13,  12,   0) =  0
	Correct value 1.959348e-03
	Computed value 1.992668e-04
	Difference 1.760082e-03
	Maximum absolute difference = 1.760082e-03
FAILED

If I run this TCL on the Python test case output, visual inspection of output shows the TCL seems to be correctly identifying the entries with the large differences.

lappend auto_path $env(PARFLOW_DIR)/bin 
package require parflow
namespace import Parflow::*

source ../tcl/pftest.tcl

set i 10

if ![pftestFile  tilted_v_with_reservoir_overland_flow.out.press.00010.pfb "Max difference in Pressure for timestep $i" $sig_digits] {
    puts "FAILED"
}

smithsg84:
The python error checking was incorrect, this is fixed in #549

smithsg84:
@westb2 Can you look at #549 which I believe fixes this issue.   The arguments to the abs file check method were inverted, causing the test to not work.   The error you found in the pressure was found and I updated with your new pressure from this PR and everything is passing.   Another test had errors so I believe #549 is now working as intended.    If you agree I'll merge #549 and close this issue.

smithsg84:
Fixed in #549

PR #557, opened 2024-10-17
Check parflow inputs for NaNs

gartavanis:
This PR adds a new method check_nans() to the Run class of pftools. This checks the static and forcing files for NaNs. The aim is to expand the suite with other input checks, e.g. mask checks, values that are out-of-range etc, depending on the type of simulation (box or solid domain). This can eventually be integrated into Run.run(), so that the checks are executed automatically before the simulation starts.

Also, the LW test.py test is modified to check its inputs.


pftools/python/parflow/tools/core.py
```diff
@@ -512,6 +514,39 @@ def run(self, working_directory=None, skip_validation=False):
         if not success or error_count > 0:
             sys.exit(1)
 
+    def check_nans(self, working_directory, exclude_forcing=True):
```
smithsg84, line 517: An API best practice I picked up somewhere is to avoid negative flags.   Suggest "include_forcing".

Generally makes the code a bit easier to read ("if include_forcing" without the 'not') and documentation reads generally reads easier "If set to True, forcing files will be checked" if you don't have to do a mental true/false inversion.

test/python/washita/LW_Test.py
```diff
@@ -561,4 +565,21 @@
...
 # -----------------------------------------------------------------------------
 
+# Corrupt the indicator file and run check_nans()
+corrupted_indicator = read_pfb(os.path.join(dir_name, "IndicatorFile_Gleeson.50z.pfb"))
+corrupted_indicator[0, 0, 0] = np.nan
+write_pfb(os.path.join(dir_name, "corrupted_indicator.pfb"), corrupted_indicator)
+LW_Test.Geom.indi_input.FileName = "corrupted_indicator.pfb"
+try:
+    LW_Test.check_nans(working_directory=dir_name)
+    raise AssertionError("Expected ValueError, but none was raised.")
```
smithsg84, line 575: Would be more clear if message was less generic like the one print below.  

"ValueError was raised as expected for corrupted indicator file."

test/python/washita/LW_Test.py
```diff
@@ -561,4 +565,21 @@
...
+except ValueError as e:
+    assert "corrupted_indicator.pfb" in str(e)
+    print("ValueError was raised as expected for corrupted indicator file.")
```
smithsg84, line 578: We tend to not print a lot of successful messages to make the successful test runs not very verbose.

smithsg84 (changes requested):

pftools/python/parflow/tools/core.py
gartavanis: Thanks for the suggestion, makes sense.

smithsg84 (approved):

PR #566, opened 2024-11-19
compressible storage positive

rlsandovalp:
The subsurface storage computation previously allowed negative storage values when the pressure head was negative, potentially leading to physically unrealistic results. This issue occurred in both the Python and Tcl pftools scripts. To address this, a minimum value of zero has been imposed for compressible storage in both scripts, ensuring that negative storage values are avoided. The implementation is backward compatible and should not affect existing workflows.

@lecondon @smithsg84 @gartavanis


smithsg84 (approved):
Thanks for patch.

PR #568, opened 2024-11-22
MacOS github actions CI fix

smithsg84:
Fix github CI action for MacOS; brew has pkg-config changes that are causing issues


danrosen25:
It looks like great progress.

PR #573, opened 2024-12-10
Draft: Add Umpire as the memory manager for the GPU backend.

gartavanis:


pfsimulator/parflow_exe/main.c
```diff
@@ -185,6 +185,10 @@ int main(int argc, char *argv [])
     RMM_ERR(rmmInitialize(&rmmOptions));
 #endif // PARFLOW_HAVE_RMM
 
+#ifdef PARFLOW_HAVE_UMPIRE
+    umpireInit();
```
smithsg84, line 189: Just an idea for a when we do a future refactor.   I like this umpireInit() method rather than the RMM initialize here in main.   I'm thinking we have something like a "pf_backendInit()" method and then we can move the initialize #ifdef stuff into that.

In general if we can define the backend API so we don't have #ifdef's all over I think it is much cleaner.    Also when we don't have a backend the API methods can just be null methods.

And when we go C++ would be nice to just have a Class to define the backend API and do subclassing for the different backends, use the strategy pattern.   In fact we might be able to make the back-end select-able at runtime.   I sort think alloc/free overhead is going to hide a function call overhead so I'm not sure why things in pf_devices.h are static inline.

OK sorta rambling at this point.  Definitely think there is some cleanup here to do later.   Starting to think the cleanup should happen after the C++ change since both AMPS and the backend are ideal use cases for a strategy pattern.

smithsg84 (changes requested):
This isn't compiling currently but changes look with a minor request to namespace the amps files/methods.

pfsimulator/amps/common/umpire_wrapper.cpp
```diff
@@ -0,0 +1,31 @@
+#include "umpire_wrapper.h"
```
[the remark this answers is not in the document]
gartavanis: Thanks for catching this - fixed!

smithsg84 (changes requested):
Looks good.

Can you look at the "README-GPU.md" file and update out-of-date (e.g. readme says to use RMM v.0.10).   Also if you have ideas for general cleanup of that text.

Add a comment on the CUDAHOSTCXX=mpicxx  config issue.

Add umpire/rmm flag comments.

And maybe a comment on the CUDA_ARCH flag (whatever that was).  

So we have something users can read rather than looking through the CMAKE files.   Building with CUDA is a bit complicated, I don't think we will capture all the issues but if we have something maybe that will help with questions.

pfsimulator/parflow_exe/main.c
smithsg84: Captured this in issue #589 for later investigation

PR #583, opened 2025-01-15
DRAFT: Hacking to get CUDA/Hypre running

smithsg84:


gartavanis (commented):
This PR has some useful bits and some that are no longer needed. A few comments:

- PF_HYPRE_StructVectorSetValuesPtr no longer needed (the hypre repo fixed the inconsistency in their interface)
- We should add the changes in linux.yml to the self-hosted runner so that the hypre-dependent tests can also be run.
- The different memory allocations (to managed memory if compiled with CUDA) are currently necessary to run parflow and hypre on GPUs. If we change the strategy for interfacing with hypre, these changes might not be needed. As a result, this PR might not be the best place to start from to optimize parflow-hypre performance on GPUs.
- The PR is stale and needs some work to resolve conflicts from the master branch.

PR #585, opened 2025-01-18
Feature/mixed surface bcs

reedmaxwell:
This feature adds a mixed boundary condition capability to ParFlow.  If the top boundary is set to OverlandKinematic and a pfsolid file is used, this solid file can have multiple patches with mixed boundary conditions.  One patch is assigned to the overland kinematic, and up to two more patches can be assigned to seepage face cells mixed in with the overland flow.

tagging @smithsg84 @lecondon to review.

### DIFF
--- a/docs/user_manual/keys.rst
+++ b/docs/user_manual/keys.rst
@@ -4452,6 +4452,42 @@ minimum value for the :math:`\bar{S_{f}}` used in the
    pfset Solver.OverlandKinematic.Epsilon 1E-7           ## TCL syntax
 
    <runname>.Solver.OverlandKinematic.Epsilon = 1E-7     ## Python syntax
+  
+
+Mixed Boundary Conditions  
+
+If the top boundary is set to **OverlandKinematic** and a pfsolid file is used, this solid file can have multiple patches with mixed boundary conditions.  One patch is assigned to the overland kinematic, and up to two more patches can be assigned to seepage face cells mixed in with the overland flow.
+
+*double* **Solver.OverlandKinematic.SeepageOne** -999  
+
+This key sets the first of two available patch values for mixed boundary conditions.
+
+::
+   pfset Solver.OverlandKinematic.SeepageOne 3           ## TCL syntax
+
+   <runname>.Solver.OverlandKinematic.SeepageOne = 3    ## Python syntax
+
+*double* **Solver.OverlandKinematic.SeepageTwo** -999 This key sets the second of two available patch values for mixed boundary conditions.
+
+::
+
+   pfset Solver.OverlandKinematic.SeepageTwo 4           ## TCL syntax
+
+   <runname>.Solver.OverlandKinematic.SeepageTwo = 4    ## Python syntax
+
+Example of mixed boundary conditions: 
+
+::
+   CONUS2.Geom.domain.Patches = "ocean land top lake sink bottom"
+   CONUS2.BCPressure.PatchNames ="ocean land top lake sink bottom"
+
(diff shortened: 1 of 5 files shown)


docs/user_manual/keys.rst
```diff
@@ -4452,6 +4452,42 @@ minimum value for the :math:`\bar{S_{f}}` used in the
...
+Mixed Boundary Conditions  
+
+If the top boundary is set to **OverlandKinematic** and a pfsolid file is used, this solid file can have multiple patches with mixed boundary conditions.  One patch is assigned to the overland kinematic, and up to two more patches can be assigned to seepage face cells mixed in with the overland flow.
+
+*double* **Solver.OverlandKinematic.SeepageOne** -999  
+
+This key sets the first of two available patch values for mixed boundary conditions.
+
+::
+   pfset Solver.OverlandKinematic.SeepageOne 3           ## TCL syntax
```
smithsg84, line 4466: Why 2 Seepage patches

We set values based on patch names in other parts of the code (e.g. BCPressure.PatchNames).

I'm a bit confused here.   "OverlandKinematic" is set for a patch.   E.G.

pfset Patch.**z-upper**.BCPressure.Type		      OverlandKinematic

So are the two other patches done with something like 

pfset Patch.**z-upper-seepage-one**.BCPressure.Type		      Seepage

pfset Patch.**z-upper-seepage-two**.BCPressure.Type		      Seepage

I'm a confused by the "mixed in" comment ,patches are non-overlapping.

Example input file might help my understanding.

pfsimulator/parflow_lib/nl_function_eval.c
```diff
@@ -41,6 +41,8 @@ typedef struct {
   double SpinupDampP1;      // NBE
   double SpinupDampP2;      // NBE
   int tfgupwind;           //@RMM added for TFG formulation switch
+  int seepage_patch_one;  //RMM added two optional seepage face BCs
```
smithsg84, line 44: Is this value and int or a double.   Seems to be read as a double then converted to int?

GetDoubleDefault is used to read and the type in the keys.rst document is double.

smithsg84 (commented):
@reedmaxwell 

I'm confused on the input setup.   I don't get the use of two seepage keys to specify the patches. See comments in PR.

smithsg84:
Feature addition needs a test case.

gartavanis:
A few more comments here:
1. I think part of the confusion in the discussion comes from the "mixed bcs" terminology. Perhaps in the docs, we can explain that patches are non-overlapping and that the "mixed bcs" refers to patches where there is a distinct boundary condition that looks both like overland_kin and seepage_face. Would that be a good description?
2. There is quite a bit of commented out code. Is this supposed to be merged as well for referencing later, or should we remove it before merging?
3. In the docs, should the SeepageOne and SeepageTwo values be set to 4 and 5 respectively? If the patches are "ocean land top lake sink bottom", sink and bottom are patch 4 and 5 if we start counting from zero. Or maybe I'm missing something.

smithsg84:
@gartavanis I refactored the seepage methods into a seperate file to reduce the code duplication.   Also changed the isseepage methods a little to avoid doing all the linear searches; turned it into a lookup table to make the method constant time.   This will cost more memory but O(1) vs O(N) lookup.   

Can you look over and see if it looks OK.  Thanks.

gartavanis:
@smithsg84 The changes look good. Thanks for the optimization with the lookup table. I built parflow again with your changes and run the 24h CONUS2 test, and it produces identical results as before.

PR #613, opened 2025-07-31
Modernize python build system

gartavanis:


gartavanis:
@smithsg84 I would keep them for now, for backwards compatibility in case someone tries to install with an older version of pip. We can add an issue to remove them in the next release or two?

smithsg84:
@gartavanis 

Looking at some of our python dependencies.   

PyYAML version 6.0.1 which was released on July 17, 2023
Imageio version 2.9.0 was released on July 6, 2020.

pip version 21.3 is needed for pyproject.toml support and was released on October 11, 2021

I think we should deprecate the old package setup so we have a single set of files to maintain.   The existing set has a number of inconsistent version info already.

I'll push up a minor set of fixes to version differences I found.

smithsg84 (approved):

PR #631, opened 2025-08-28
Fix dependencies in pyproject.toml

gartavanis:
With the new pftools build, the dependencies like numpy and pandas weren't pulled automatically. This PR fixes this. We also need to publish a new pftools version on PyPI so that this is available to users.


smithsg84:
The pftools version number needs to be incremented for the PyPI publush.

smithsg84 (approved):

Issue #643, opened 2025-09-19
default_single.py does not call ParFlow with multiple processes


tikoh-station:
The python script `default_single.py` is called with one process for every processor topology.
It should be reading the system arguments to do so.

https://github.com/parflow/parflow/blob/49a7d7f1b4364f7fd3eeec228ae9b0669cc1a3d9/test/python/default_single.py#L23-L25

PR #644, opened 2025-09-19
Fix process topology of python tests

tikoh-station:
This PR addresses Issue #643:
- Some python tests were running with 2D or 3D topologies while setting (P, Q, R) = (1, 1, 1) in the input script;
- The python tests where this happened (including `default_single.py`) were identified;
- This PR fixes the test scripts and ensures they are running the correct topologies.


smithsg84 (approved):
Thanks for catching this bug!

PR #654, opened 2025-10-12
Add qx_overland and qy_overland surface flow output

reedmaxwell:
I added the overland flow to the Surface Predictor capability.  Along the way I realized it was easier to grab the surface flux values from nl_function_eval the same way the velocities are grabbed (the surface predictor also uses the velocities to calculate ponding).  Once I had done that I figured this was a useful thing to be able to write out, I confirmed that the output cell fluxes match the PFTools Python Hydrology Package.  This was collaborative with Claude Code over part of a weekend, I will export the transcript and share if you are interested.  I've double checked all the code additions.

- Add q_overlnd_x and q_overlnd_y vectors to store surface flow velocities
- Implement output for all formats (PFB, NetCDF, SILO, PDI)
- Add 8 new output keys to key database and user manual
- Populate qx/qy arrays in all three overland flow methods:
  * OverlandFlow (OverlandBC)
  * OverlandKinematic (OverlandKinematicBC)
  * OverlandDiffusive (OverlandDiffusiveBC)
- Update surface predictor to use qx/qy arrays for lateral flux calculations
- Values represent surface flow velocity (ke_, kn_) in m/hr
- Cell-centered for OverlandFlow, edge-centered for Kinematic/Diffusive

🤖 Generated with [Claude Code](https://claude.ai/code)


reedmaxwell:
@smithsg84 I added one test /test/python/overland_flux_test.py which checks the new output against correct output.  This test should pass now.  We can keep the other tests in the test directory but should only use this new one and associated output for the CI test suite.  thanks!

smithsg84:
> @smithsg84 I added one test /test/python/overland_flux_test.py which checks the new output against correct output. This test should pass now. We can keep the other tests in the test directory but should only use this new one and associated output for the CI test suite. thanks!

Fixed some things:

- The overland_flux test was not in the CMakeLists.txt file so was not being run with the test suite.   
- Moved the tests that are not being used to a "not-working" subdirectory to make it more clear they are not working.
- Did some minor name/cleanup on the test case.

Issue #588, opened 2025-01-29
CLM code duplication


smithsg84:
CLM is duplicating  the top and a bottom calculation.   PF has a top and bottom 2D array with the z index of top/bottom.

Vector *top = ProblemDataIndexOfDomainTop(problem_data);

The CLM duplication can be eliminated by passing top/bottom into CLM.

Depends on #547 for bottom vector.

PR #656, opened 2025-10-17
Add a new CoLM option for LSM

aureliayang:
Users can use the following key to switch between LSMs       
xxxx.Solver.LSM  = 'CoLM'

There are two examples in _pfsimulator/colm/examples_
Once you installed this version parflow
set **xxxx.Solver.LSM  = 'CoLM'** and run the colm_version 
set **xxxx.Solver.LSM  = 'CLM'** and run the clm_version


andrewsoong:
What is the version of the CLM module in Parflow?
CLM3.5 or CLM5.0？

pfsimulator/colm/main/CoLM.F90
```diff
@@ -0,0 +1,958 @@
...
+                     topo_mask(1,numpatch) = k
+                     planar_mask(1,numpatch) = i
+                     planar_mask(2,numpatch) = j
+                     planar_mask(3,numpatch) = 1
+                  end if
+               endif
+   
+               if (topo(l) == 0 .and. topo(l+k_incr) > 0) topo_mask(3,numpatch) = k + 1
+   
+            enddo ! k
```
tikoh-station: This part of the code would reintroduce Issue #588.
The computation of the bottom index of the domain has been added to ParFlow, so you don't have to recompute it inside CoLM. Check how it was fixed for CLM:
https://github.com/parflow/parflow/blob/515956a2c9b81a05e680b823ffd571369d24d794/pfsimulator/clm/clm.F90#L336-L340
Instead of passing the topo to CoLM, you can just pass the top and bottom index of the domain.

Also, this reintroduction of the issue makes me wonder: how much more overlap is there between the two models CLM and CoLM? At a quick scan, I can't find any more glaring instances where this happens.
Is there a way to group common functionalities?

tikoh-station (changes requested):
Thank you for contributing to ParFlow!
I made a very brief high-level review of your pull-request.

Given the current state of the PR, I am requesting some changes before a more detailed review can be done:
- As pointed out in the code comments, do not reintroduce Issue #588.
- This PR is missing automated CI tests that need to be added with every new feature. Given that this is another land surface model, is has to come with its own comprehensive set of tests.
- This PR is missing documentation for both the input keys and documentation about the model itself. From the example case, it seems like it repurposes CLM keys. That should be stated somewhere in the documentation too.
- Discussion: should CoLM have its own compiler flag?

aureliayang:
Dear colleagues,

Together with my students @TangZey (Zeyu Tang) and @HH12138666 (Hui Huang), we have carefully worked through the review comments and revised the PR accordingly.

@tikoh-station Thank you very much for the high-level review. We have updated the PR to address the requested changes.

**1. Issue #588 / top-bottom indexing:**

- The CoLM interface has been updated to use the ParFlow-provided top and bottom indices of the active domain, following the same approach used by CLM. CoLM no longer recomputes the active-domain bottom index from the topo mask, and the extra variables that were only needed for that recomputation have been removed.

**2. CI tests:**

- Automated CI tests have been added for the CoLM configuration and example case. These tests are included in the GitHub Actions workflow so that the CoLM coupling is built and checked together with the existing ParFlow test infrastructure.

**3. Documentation, input keys, and examples:**

- The ParFlow documentation now includes CoLM in the standard land-surface-model workflow. The `Solver.LSM` key documentation lists `CoLM` as a supported runtime option alongside `none` and `CLM`, and the manual introduction/start section describes CoLM together with CLM as a land-surface coupling option.

- The key definitions were updated in `pf-keys/definitions/solver.yaml`, and the corresponding user-manual key documentation in `docs/user_manual/keys.rst` now reflects that `Solver.LSM` supports `none`, `CLM`, and `CoLM`. Because the current CoLM interface intentionally reuses the existing `Solver.CLM.*` namespace for shared land-model coupling controls, this is stated explicitly in the CoLM documentation instead of introducing a parallel set of CoLM keys.

- A runnable CoLM example is provided in `pfsimulator/colm/examples/colm_version`. The dedicated `README-CoLM.md` provides the detailed CoLM-specific documentation, including how to enable CoLM, how the existing `Solver.CLM.*` keys are used, which keys are active versus retained for compatibility, and which additional CoLM input files are required by the example.

**4. Compiler flag discussion:**

- Our current preference is to keep CoLM under the existing `PARFLOW_HAVE_CLM` build option and select the active land surface model at runtime using `Solver.LSM`. The reason is that CoLM currently reuses much of the existing CLM coupling infrastructure, including the build configuration, CMake wiring, and solver-side integration. Introducing a separate CoLM-specific build flag at this stage would require splitting or duplicating a substantial amount of shared logic, which may make the code harder to maintain.

- From the user side, a single build with runtime selection through `Solver.LSM = "CLM"` or `Solver.LSM = "CoLM"` is also simpler and avoids requiring users to rebuild ParFlow when switching between the two land surface models. This has also been our experience in recent ParFlow training, where users found runtime selection through an input key more convenient than managing separate build configurations.

We believe the requested review comments have now been addressed and the PR should be ready for the next stage. We would be happy to discuss any further questions :)

PR #657, opened 2025-10-17
Add option to use Sundials library instead of the embedded Kinsol

smithsg84:
The embedded Kinsol version is very old and has not been updated since the early versions of ParFlow.   A compile time option has been added to use the Sundials library version of Kinsol.

Some tests were modified to use EtaConst as this is the more stable option but possibly less performant.   This provided consistency between the old and new Kinsol results.

Some tests were setting EtaValue when using Walker1 and Walker2 EtaChoice options.   The key is ignored for those options so removed for better clarity.

A bug in Walker1 was backported to the embedded Kinsol version from the new Sundials.   This provides better consistency between the old and new Kinsol and should provide better solver performance for Walker1.

Walker2 results were the most inconsistent between the old and new kinsol.   Setting EtaGammer to a tighter constraint improves consistency.

### DIFF
--- a/pftools/pfmask-to-pfsol.cpp
+++ b/pftools/pfmask-to-pfsol.cpp
@@ -417,7 +417,7 @@ int main(int argc, char **argv)
   double sx = 0, sy = 0, sz = 0;
   double dz;
 
-  std::vector<Databox*> databox(inFilenames[0].size());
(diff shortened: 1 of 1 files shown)


oseikuffuor1 (approved):
Looks good Steve. Thanks!

PR #659, opened 2025-10-19
Fix elai and esai being overwritten

aureliayang:
when **clm_forc_veg== 1**, we read _elai_ and _esai_ from files and setup in **subroutine drv_getforce** as follows:
```
	if  (clm_forc_veg== 1) then 
		clm(t)%elai	        = lai_pf(l)
		clm(t)%esai	        = sai_pf(l)	
		clm(t)%z0m	        = z0m_pf(l) 
		clm(t)%displa	    = displa_pf(l)     
	endif
```
However, these values will be overwritten in **subroutine clm_dynvegpar** as
```
      clm%elai = clm%tlai*(1.-fb)
      clm%esai = clm%tsai*(1.-fb)
```

I fix this issue by adding if condition:
```
      if  (clm_forc_veg == 0) then
          clm%elai = clm%tlai*(1.-fb)
          clm%esai = clm%tsai*(1.-fb)
      endif
```

Reed, please help to review this issue @reedmaxwell , thanks much!


reedmaxwell:
I reviewed and I think this looks good to me.  If the CLM tests still pass then the internal LAI/SAI function should be working correctly.

PR #669, opened 2025-11-09
[Enhancement] Improve Umpire Allocator Selection

mfahdaz:
**Background**

Previously, ParFlow configured Umpire to always use the `UM_pool `(Unified Memory) allocator.
While this performs well on NVIDIA GPUs, where Unified Memory provides efficient page migration and mature driver support, it performs poorly on AMD MI250X (ROCm) systems.

ROCm’s managed memory (`hipMallocManaged`) still incurs higher paging and migration overheads, particularly when Heterogeneous Memory Management (HMM) or XNACK is disabled. In these configurations, managed allocations fall back to pinned host memory, resulting in slow device access and degraded solver performance.

**Observations**: When running ParFlow with Kokkos + HIP + Umpire (`UM_pool`) on AMD MI250X GPUs, initialization completes quickly, but the KINSOL solver progresses significantly more slowly, leading to substantial runtime degradation. **NOTE**: This behavior is related to performance, not numerical convergence or solver correctness.

**Summary of Changes**

This PR refactors Umpire initialization in ParFlow to dynamically select the most appropriate allocator based on the active Kokkos backend, improving both performance and portability across GPU architectures.

**Key Updates**

- Allocator Selection Logic
  - `UM_pool` used for CUDA (native or Kokkos), SYCL, and OpenMP (Kokkos) backends
  - `DEVICE_pool` used for HIP (Kokkos) backends
  - `HOST_pool` used as a CPU fallback
- Backend Detection
  - Updated CMake configuration to propagate backend macros:
    - KOKKOS_ENABLE_CUDA
    - KOKKOS_ENABLE_HIP
    - KOKKOS_ENABLE_SYCL
    - KOKKOS_ENABLE_OPENMP_TARGET
- CMake and Logging Cleanup
  - Simplified Umpire/RMM messaging now emits a single status line showing the selected backend and memory manager.
  - Log files now report the accelerator backend and the selected memory pool allocator.
- Expanded Backend Coverage
  - Unified logic for native CUDA, Kokkos-enabled CUDA, HIP, SYCL, and OpenMP Target backends.
  
**Performance Impact**
This update significantly improves GPU performance and portability:

- **AMD MI250X (Kokkos (HIP) )**
 Switching from `UM_pool` to `DEVICE_pool` restores expected solver performance, reducing solvetime to ~25% and total runtime by ~50% compared to runs without Umpire (i.e., Kokkos(HIP) only).
- **NVIDIA A100 (CUDA / Kokkos (CUDA) )** 
No regressions observed and performance remains consistent with the previous benchmarks.

**Testing & Verification**

Validated configurations:

- Native CUDA (A100)
- Kokkos-enabled CUDA (A100)
- Kokkos-enabled HIP (MI250X)

Runtime validation confirms:

- Correct allocator selection for each backend
- Stable solver convergence across all configurations
- Expected runtime improvements on AMD GPUs without performance loss on NVIDIA

### DIFF
--- a/pfsimulator/amps/common/amps_umpire_wrapper.cpp
+++ b/pfsimulator/amps/common/amps_umpire_wrapper.cpp
@@ -9,66 +9,71 @@
 
 extern "C" {
 // -----------------------------------------------------------------------------
-// Helper function: determine the correct base allocator name based on backend.
+// Helper: Select correct Umpire memory resource based on active backend
 // -----------------------------------------------------------------------------
-static std::string get_base_allocator_name()
+static std::string get_resource_name()
 {
 #if defined(PARFLOW_HAVE_CUDA) || defined(KOKKOS_ENABLE_CUDA)
-  // Native CUDA or Kokkos-enabled CUDA
-  return "UM";  // Unified Memory
+  return "UM";
 #elif defined(KOKKOS_ENABLE_HIP)
-  // Kokkos-enabled HIP (AMD)
   return "DEVICE";
 #elif defined(KOKKOS_ENABLE_SYCL)
-  // Kokkos-enabled SYCL (Intel/oneAPI or other SYCL devices)
   return "UM";
 #elif defined(KOKKOS_ENABLE_OPENMP_TARGET)
-  // OpenMP offload targets (NVIDIA/AMD/Intel GPUs)
   return "UM";
-#else
-  // CPU-only fallback
+#else  // CPU-only fallback
   return "HOST";
 #endif
 }
 
 // -----------------------------------------------------------------------------
-// Initialize Umpire allocators and dynamic memory pools
+// Helper: Name of the pooled allocator constructed from the resource
 // -----------------------------------------------------------------------------
-void amps_umpireInit()
+static std::string get_pool_name()
 {
-  // Create a pool allocator
-  // Initial size (default: 512MB) and pool increase size (1MB) can be tuned in the constructor.
-  auto& rm = umpire::ResourceManager::getInstance();
-
-  std::string base_allocator_name = get_base_allocator_name();
-  auto base_alloc = rm.getAllocator(base_allocator_name);
-  auto pooled_alloc =
-    rm.makeAllocator<umpire::strategy::DynamicPoolList>(
-                                                        base_allocator_name + "_pool", base_alloc);
+  return get_resource_name() + "_pool";
 }
 
 // -----------------------------------------------------------------------------
-// Allocate memory using the correct pooled allocator
+// Initialize Umpire allocators and pool strategy.
+// -----------------------------------------------------------------------------
+void amps_umpireInit()
+{
+  auto& rm = umpire::ResourceManager::getInstance();
+  auto resource_name = get_resource_name();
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -74,15 +75,39 @@ elseif(PARFLOW_ACCELERATOR_BACKEND STREQUAL "cuda")
...
+    elseif(Kokkos_ENABLE_OPENMP_TARGET)
+      message(STATUS "Kokkos-enabled OpenMP target-offload backend selected")
+      set(KOKKOS_ENABLE_OPENMP_TARGET "yes")
+
+    else()
+      message(STATUS "Kokkos host-only backend selected")
+      set(KOKKOS_ENABLE_SERIAL "yes")
+    endif()
+
+  else()
```
gartavanis: Same for this else().

pfsimulator/amps/common/amps_umpire_wrapper.cpp
```diff
@@ -7,25 +8,70 @@
...
+#elif defined(KOKKOS_ENABLE_SYCL)
+  // Kokkos-enabled SYCL (Intel/oneAPI or other SYCL devices)
+  return "UM";
+#elif defined(KOKKOS_ENABLE_OPENMP_TARGET)
+  // OpenMP offload targets (NVIDIA/AMD/Intel GPUs)
+  return "UM";
+#else
+  // CPU-only fallback
+  return "HOST";
+#endif
```
gartavanis: I think it makes more sense to have the full names here ("UM_pool" etc). The code will be clearer, and if we want to add more allocators (which might not be pool allocators) we will only need to add them to this function without modifying the init, alloca and free functions.

gartavanis (commented):
Thanks for adding this feature! The PR looks good overall, I just had minor comments.

Regarding the PR summary: I think it should be clearer if the reason for adding this feature is performance or accuracy. Based on the proposed changes I think it is performance, but the summary is a little confusing as it mentions "the KINSOL solver converges significantly more slowly" and then later, "Stable solver convergence across all configurations". When I first read this, I thought that the solver was solving the problem in a different way. I would clarify that this is not about solver convergence, but solver performance.

mfahdaz:
> Thanks for adding this feature! The PR looks good overall, I just had minor comments.
> 
> Regarding the PR summary: I think it should be clearer if the reason for adding this feature is performance or accuracy. Based on the proposed changes I think it is performance, but the summary is a little confusing as it mentions "the KINSOL solver converges significantly more slowly" and then later, "Stable solver convergence across all configurations". When I first read this, I thought that the solver was solving the problem in a different way. I would clarify that this is not about solver convergence, but solver performance.

Thank you for the feedback! Your comments really helped improve the overall clarity.

Yes, this PR primarily focuses on performance improvement on AMD GPUs, not a change in solver accuracy or numerical convergence. To clarify: on AMD GPUs, when using `UM_pool`, the solver performs all iterations but progresses significantly more slowly. Switching to `DEVICE_pool` restores expected performance, with the solver completing significantly faster than even the baseline without Umpire. I’ve updated the PR summary to make this distinction clearer.

PR #699, opened 2026-02-06
Hypre logging

gartavanis:
Fixes Hypre logging segfaults from inside ParFlow. With those changes, the hypre norms will be printed on the out.log parflow file. Not sure if we always want to do this - perhaps we should put guards around the logging like the old `tol > 0.0` guards. But it should be something that can be turned on and off by the user, unlike the old `tol` guard.


smithsg84:
> ?

If there isn't a performance issues (I didn't see one), making it a runtime option and just using a new input key to enable/disable would make it easier for it to be turned on/off.

PR #742, opened 2026-05-28
Linux binaries

gartavanis:


smithsg84:
@gartavanis Was using an existing like CPACK looked at?   I'm not sure how that would work if you are attempting to bundle the MPI executables so maybe not an option.   CPACK supports building standard distribution files, tarballs, self extracting shell scripts, RPM/DEB etc.

And MAC.

gartavanis:
@smithsg84 I wasn't sure how we could bundle together all dependencies (MPI, HYPRE, NETCDF, etc) with CPACK and create a relocatable executable. We can look into this more.

smithsg84:
> @smithsg84 I wasn't sure how we could bundle together all dependencies (MPI, HYPRE, NETCDF, etc) with CPACK and create a relocatable executable. We can look into this more.

I've not tried it but I believe bundling dependencies is supported.

https://discourse.cmake.org/t/how-to-cpack-dependent-shared-libraries-along-with-my-executable/267?utm_source=chatgpt.com

I'm not sure about MPI executables (mpiexec) for running.

smithsg84 (approved):

PR #744, opened 2026-05-30
Macos binaries

gartavanis:
Build binaries that can be downloaded and run on macOS.


README-MACOS-BINARY.md
```diff
@@ -0,0 +1,75 @@
...
+
+```bash
+# 1. Download and extract (substitute the actual version)
+curl -LO https://github.com/parflow/parflow/releases/download/vX.Y.Z/parflow-vX.Y.Z-macos-arm64.tar.gz
+tar xzf parflow-vX.Y.Z-macos-arm64.tar.gz
+
+# 2. Remove macOS quarantine flag
+xattr -dr com.apple.quarantine install/
+
+# 3. Set environment (add to your shell profile for persistence)
```
smithsg84, line 18: Why is this different from sourcing a script method used for Linux binaries?

README-MACOS-BINARY.md
```diff
@@ -0,0 +1,75 @@
...
+export PARFLOW_DIR=$PWD/install
```
smithsg84, line 19: The comment says to add this to the shell profile but $PWD would be home directory.   If people download to someplace other than home dir putting this line in profile won't work.   The Linux install is more explicit.

cd /path/to/release-install
source parflow-env.sh

I think Linux and MacOS install should mirror each other for simplicity.

smithsg84 (changes requested):

smithsg84 (approved):

PR #746, opened 2026-06-12
DRAFT: Interface with ALQUMIA

smithsg84:


smithsg84:
Some things that need work

- [x] Should we move the interfacing code to a seperate directory to make code structure cleaner?
- [x] All new functions should have doxygen comments added for the function; we are requiring documentation now
- [ ] We want to move to C++ for compiling so the C restriction should be removed.  Normally fixable with some extern C additions around the C API.
- [x] Should look at why we need a Global.   We should try to avoid introducing additional global state.  Maybe put flag in problem data structure?  Guess we need to see where it is used.
- [x] advect_new.F90 is not a good name.   advect_alquimia?  or is it more generic?
- [x] Should we change Solver.Chemistry to be a Name rather True/False flag if other options are a potential?
- [ ] There a similar code blocks added to the solvers, an these blocks be extracted to functions to reduce duplication and reduce the ever growing size of the solver code?

reedmaxwell:
@smithsg84 great points, thanks for the quick review, two of your points addressed already, others planned in the next few iterations.

On #4 (Global state → Problem):
Done. Chemistry state no longer lives in globals — it's carried in the Problem data structure and accessed via ProblemChemistry(problem) throughout, with ProblemChemistryFromInput() populating it at parse time. The old global was deleted. (commit 062c7f9c)

On #6 (Solver.Chemistry as a Name):
Done. Solver.Chemistry is now a Name key with enum {None, Alquimia} rather than a True/False flag, so additional engines can be added as new names without another boolean. (commit 062c7f9c)

reedmaxwell:
@smithsg84:
per your comments, 
* moved chem_* into pfsimulator/parflow_lib/chemistry/
* renamed advect_new.F90 -> advect_transport.F90
* doxygen headers added
* differences in solver transport blocks make dedup tricky (see longer comment)

reedmaxwell:
## Summary

This PR ports the abandoned `react_trans` branch's reactive-transport coupling
onto current master: ParFlow solute transport coupled to a geochemical engine
(CrunchFlow / PFLOTRAN) via the **Alquimia v1.0** interface using explicit
operator splitting. Transport and chemistry run on CPU (v1; GPU out of scope).

**Default-off and behavior-neutral.** `Solver.Chemistry = None` by default; with
chemistry off the code path is identical to master and the full test suite
passes with zero baseline diffs. Chem sources are excluded from the build when
Alquimia is off and never reach the GPU build; a configure guard errors on
Alquimia + CUDA/Kokkos.

**How it works.** Each step advects the mobile concentrations (2nd-order Godunov,
no reaction term), then the engine advances geochemistry per cell over the same
interval (`ReactionStepOperatorSplit`); the sub-step is CFL-limited. The
geochemical system is given in a native engine deck named by `Chemistry.*`;
initial/boundary chemistry via `GeochemCondition.*` / `BCConcentration.*`.

**Verification.** The 1D `calcite` and `tracer` acceptance-gate cases reproduce
Molins et al. (2025, GMD 18:3241). All six Amanzi chemistry benchmark cases
(tracer, calcite, tritium, ion exchange, surface complexation, isotherms) were
cross-checked against the shipped PFLOTRAN reference for the identical decks:
mobile-species/pH front positions agree to ≈1 grid cell or better, conservative
species to ~0.1 m, and sorbed concentrations to ~0.1 %. The port's fronts are
sharper than PFLOTRAN's because ParFlow's advection applies no molecular
diffusion (the engine deck's `fix_diffusion` is not used) — a modeling
difference, not a coupling error.

**Included:** the coupling + both solvers (Richards, IMPES); calcite/tracer
tests (TCL + Python) behind `PARFLOW_HAVE_ALQUIMIA`; pf-keys definitions; a
chemistry section + key reference + annotated example in the user manual; a
reproducible engine-build script; doxygen on the new functions; chem sources
relocated to `parflow_lib/chemistry/`.

**Open questions for reviewers:**
1. **Solver-block dedup** — the genuine duplication is between `impes` and `lb`
   (rarely/never used); `richards` (production) is structurally different. Worth
   only a scoped extraction of the shared per-contaminant kernel? (details in
   thread)
2. **Compile-as-C++ / `extern "C"`** — this reverses a v1 hard constraint; OK to
   land in this PR or defer?
3. The four extra benchmark cases beyond calcite/tracer are verified but not yet
   committed — add as CTest regression tests here, or keep as a separate
   benchmark suite?

reedmaxwell:
## Updated Summary

This PR ports the abandoned `react_trans` branch's reactive-transport coupling
onto current master: ParFlow solute transport coupled to a geochemical engine
(CrunchFlow / PFLOTRAN) via the **Alquimia v1.0** interface using explicit
operator splitting. Transport and chemistry run on CPU (v1; GPU out of scope).

**Default-off and behavior-neutral.** `Solver.Chemistry = None` by default; with
chemistry off the code path is identical to master and the full test suite
passes with zero baseline diffs. Chem sources are excluded from the build when
Alquimia is off and never reach the GPU build; a configure guard errors on
Alquimia + CUDA/Kokkos.

**How it works.** Each step advects the mobile concentrations (explicit
operator-split flux-corrected-transport scheme, no reaction term), then the
engine advances geochemistry per cell over the same
interval (`ReactionStepOperatorSplit`); the sub-step is CFL-limited. The
geochemical system is given in a native engine deck named by `Chemistry.*`;
initial/boundary chemistry via `GeochemCondition.*` / `BCConcentration.*`.

**Verification.** The 1D `calcite` and `tracer` acceptance-gate cases reproduce
Molins et al. (2025, GMD 18:3241). All six Amanzi chemistry benchmark cases
(tracer, calcite, tritium, ion exchange, surface complexation, isotherms) were
cross-checked against the shipped PFLOTRAN reference for the identical decks:
mobile-species/pH front positions agree to ≈1 grid cell or better, conservative
species to ~0.1 m, and sorbed concentrations to ~0.1 %. The port's fronts are
sharper than PFLOTRAN's because ParFlow's advection applies no molecular
diffusion (the engine deck's `fix_diffusion` is not used) — a modeling
difference, not a coupling error.  These tests running locally, should be included somewhere but perhaps not as part of CI suite (see below).

**Included:** the coupling + both solvers (Richards, IMPES); calcite/tracer
tests (TCL + Python) behind `PARFLOW_HAVE_ALQUIMIA`; pf-keys definitions; a
chemistry section + key reference + annotated example in the user manual; a
reproducible engine-build script; doxygen on the new functions; chem sources
relocated to `parflow_lib/chemistry/`.

**Open questions for reviewers:**
1. **Solver-block dedup** — the genuine duplication is between `impes` (rarely used) and `lb`
   (never used); `richards` (production) is structurally different. Worth
   only a scoped extraction of the shared per-contaminant kernel? (details in
   thread)
2. **Compile-as-C++ / `extern "C"`** — LMK if you want that included
3. The four extra benchmark cases beyond calcite/tracer are verified but not yet
   committed — add as CTest regression tests here, or keep as a separate
   benchmark suite?

PR #751, opened 2026-07-22
Fix CLM restart read/write asymmetry for snowage_vis/nir

reedmaxwell:
## Bug

`drv_restart.F90` wrote `snowage_vis` and `snowage_nir` (the two-band snow age fields) mid-record, immediately after `snowage`, but the read path never consumed them. Every subsequent read was shifted by two records, so any CLM restart written with these fields failed at step 0 with:

```
Fortran runtime error: I/O past end of record on unformatted file
```

Net effect: CLM restarts were unreadable on any build with the two-band snow albedo restart variables.

## Fix

Move both writes to the appended-new-fields block at the end of the file, after `coszen_avg`, and read all three there with `iostat` guards. This is the same backward-compatible pattern `coszen_avg` already uses.

## Compatibility

- Old restart files that predate these fields remain readable: the trailing reads fail cleanly, `coszen_avg` defaults to 0, and `snowage_vis`/`snowage_nir` fall back to the legacy scalar `snowage` (the pre-two-band behavior).
- Restart files written by the broken code were never readable by any binary, so no working files are orphaned by the layout change.

## Validation

- Watershed test (East River, CONUS2.1 subset, WY2017): a year-2 restart resumes exactly at year 1's end state (`t_soisno` carries), and year 1 re-runs byte-identical, confirming the restart write is output-only and does not perturb the trajectory.
- Round-trip check on the standard `test/python/clm` box case: run A writes restarts, run B (`clm_ic 1`) reads them back and runs; the restart file's trailing records are `coszen_avg`, `snowage_vis`, `snowage_nir` in order, and no `iostat` fallback fires on new-format files.

### DIFF
--- a/test/python/clm/clm_restart.py
+++ b/test/python/clm/clm_restart.py
@@ -142,9 +142,7 @@ def check_restart_structure(path):
         ok = False
 
     # The three appended fields must hold finite, non-negative values
-    for name, rec in zip(
-        ("coszen_avg", "snowage_vis", "snowage_nir"), records[-3:]
-    ):
+    for name, rec in zip(("coszen_avg", "snowage_vis", "snowage_nir"), records[-3:]):
         values = np.frombuffer(rec, dtype=np.float64)
         if not np.all(np.isfinite(values)) or np.any(values < 0.0):
             print(f"FAILED: restart field {name} has bad values: {values}")
--- a/test/python/clm/clm_restart.py
+++ b/test/python/clm/clm_restart.py
@@ -1,0 +1,476 @@
+# -----------------------------------------------------------------------------
+# CLM restart round-trip test
+#
+# Guards the write/read symmetry of the CLM restart file
+# (pfsimulator/clm/drv_restart.F90). Every field written by the rw=2 block
+# must be consumed by the rw=1 block at the same position in the file. A
+# field written mid-record without a matching read shifts every later record,
+# and the restarted run dies at step 0 with "Fortran runtime error: I/O past
+# end of record". This happened when snowage_vis/snowage_nir were added:
+# they were written after snowage but never read back.
+#
+# Three runs on the standard 5x5 CLM box with snow forcing:
+#   cold : hours 0-5, cold start, writes a CLM restart file every step
+#   cont : hours 0-10, cold start, reference trajectory
+#   warm : hours 5-10, restarted from cold's step-5 restart file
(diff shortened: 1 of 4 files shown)


gartavanis (commented):
The code looks good. The PR summary mentions validation with two tests, but none are included in this PR. I think it would be good to add at least one test guarding against this bug (write an rst file with all fields and read it back), as more fields added in the future could cause the same issue.

As part of the parflow manual update, in the section "Contributing new keys" we can also include a subsection about adding new clm variables and updating the above test case to ensure consistency.

reedmaxwell:
thanks George (@gartavanis), good catch.  Sorry that got missed the first time.

Test (test/python/clm/clm_restart.py, registered as py_clm_restart): runs the standard 5x5 CLM box three ways — a cold start writing a restart every step, a continuous 0–10 hr reference, and a warm run restarted from the cold run's step-5 restart file. It checks (1) the restart file's record layout against a documented table (count, order, byte size), so a field written without a matching read fails immediately, and (2) that the restarted run reproduces the continuous trajectory in pressure, saturation, and CLM output to 6 significant digits. With the original bug the warm run aborts at its first step, so both failure modes are covered. The test is self-referential (warm vs. its own continuous run), so it needs no correct_output files and won't drift across platforms.

Manual: added an "Adding new CLM variables" section to the Contributing-keys page documenting the append-at-end rule, the  iostat-guarded read with a legacy default, and the requirement to update the test's EXPECTED_RECORDS when adding a restart field.

test/python/clm/clm_restart.py
```diff
@@ -0,0 +1,474 @@
...
+            f"Max difference in warm vs continuous {label}, timestep {timestep}",
+        ):
+            passed = False
+    if not pf_test_file(
+        f"{warm_dir}/clm_restart_warm.out.clm_output.{timestep}.C.pfb",
+        f"{cont_dir}/clm_restart_cont.out.clm_output.{timestep}.C.pfb",
+        f"Max difference in warm vs continuous CLM output, timestep {timestep}",
+    ):
+        passed = False
+
```
gartavanis, line 464: Is there a reason this check is separate? I think we can add the tuple ("clm_output", "CLM output") to the inner loop above to simplify this.

gartavanis (approved):

smithsg84 (approved):

PR #752, opened 2026-07-22
Fix CLM segmentation fault for inactive ParFlow columns

TangZey:
Dear ParFlow maintainers,

@aureliayang (Chen Yang) and I found a bug in the ParFlow-CLM coupling that can cause a segmentation fault during CLM initialization.

The error was reported as:

```text
Process received signal
Signal: Segmentation fault (11)
Signal code: Address not mapped (1)
mpiexec noticed that process rank 0 exited on signal 11
```

The problem occurs when a CLM tile corresponds to an inactive ParFlow column. In that case, `top(l)` is not positive and `topo_mask(1)` is not assigned a valid top index. However, the original code continues to use `topo_mask(1)` to construct an index for reading the porosity array:

```fortran
if (top(l) > 0) then
   clm(t)%topo_mask(1) = 1+top(l)
   clm(t)%topo_mask(3) = 1+bottom(l)
   clm(t)%planar_mask = 1
endif

clm(t)%topo_mask(2) = clm(t)%topo_mask(1)-nlevsoi

do k = 1, nlevsoi
   l = 1+i + j_incr*(j) + &
       k_incr*(clm(t)%topo_mask(1)-(k-1))
   pf_porosity(k) = porosity(l)
enddo
```

This can produce an invalid memory access. The problem is environment-dependent and was not reproduced on every machine. During our testing, the same unmodified simulation ran successfully on some systems but terminated with a segmentation fault on a system built with GNU 13.

We fixed the problem by continuing to the next CLM tile when the corresponding ParFlow column is inactive:

```fortran
if (top(l) > 0) then
   clm(t)%topo_mask(1) = 1+top(l)
   clm(t)%topo_mask(3) = 1+bottom(l)
   clm(t)%planar_mask = 1
else
   cycle
endif
```

By comparing the source code across releases, we found that this invalid porosity access was introduced in ParFlow 3.14 when the porosity lookup was moved before `drv_clmini`. In ParFlow 3.13, the porosity lookup was protected by `planar_mask == 1`. The affected logic remains present in ParFlow 3.15.

We validated the fix using a watershed-scale ParFlow-CLM simulation with a 274 x 336 horizontal grid. In the GNU 13 environment, this test previously terminated with a segmentation fault during CLM initialization. After applying this change, the watershed simulation initializes and runs successfully.


smithsg84 (approved):
Thanks for the CLM bugfix

Issue #745, opened 2026-06-10
Hypre CUDA interfacing


smithsg84:
Profiling shows significant time spent in copying data to/from HYPRE structures when using CUDA.

- Get access to Princeton machines / data / problem
  - Daniel/George
- Can we subset Princeton problem, probably around coastline (assuming that # of boxes visited in Octree, many CUDA kernel launches is the issue).
- Can we setup a problem to expose Octree iteration?
  - Steve looking at setting up simple problem
- Should profile deeper
  - Hypre supports Caliper
  - Can we profile PF at finer level as well (beyond the existing PF timings)
- Should look at AI suggestions
  - Rob does not think the reuse suggestions will make much of a difference.
- Reuse the preconditioner?
  - Daniel looking at this
- What is up with the 100+ iterations in the Princeton runs for some timesteps?
  - Hypre team was interested in this.
  - Reed has thoughts
  - Is this different from previous saturation issues?
- If Octree looping / # of CUDA kernel launches is the issue
  - Can we just use single copy?
  - Currently loop was setup in CPU only days and was avoiding the inactive regions
  - Could do test after we have access to problem (or a simple problem) above.

smithsg84:
# HYPRE Interface Performance Improvements

## Summary

The current ParFlow ↔ HYPRE interface already uses HYPRE's box-based transfer routines and avoids the significantly more expensive pointwise matrix assembly path. However, there are several opportunities to improve performance, reduce maintenance burden, and improve compatibility with future HYPRE releases.

## Current State

The implementation currently:

- Uses `hypre_StructMatrixSetBoxValues()` for matrix assembly.
- Uses `hypre_StructVectorSetBoxValues()` for vector transfers.
- Avoids pointwise assembly for the bulk of matrix and vector data.
- Uses a zero-copy approach for stencil data by looping over stencil entries individually.
- Uses pointwise updates only for surface/overland-flow corrections.

This is already substantially more efficient than calling `HYPRE_StructMatrixSetValues()` or `HYPRE_StructVectorSetValues()` for every grid point.

---

## Recommendation 1: Replace Internal hypre APIs with Public HYPRE APIs

### Current

The implementation directly calls internal HYPRE routines:

```c
hypre_StructMatrixSetBoxValues(...)
hypre_StructVectorSetBoxValues(...)

These are not part of the public API and may change between HYPRE releases.

Recommended

Replace with the public interfaces:

HYPRE_StructMatrixSetBoxValues2(...)
[...]

Issue #251, opened 2020-08-05
Hypre duplicated code


smithsg84:
PFMG and SMG share common setup, matrix/vector assembly interface code for Hypre.   This should be refactored out into a set of common functions.

Issue #106, opened 2019-02-11
Hypre interface improvement


smithsg84:
The PFMG and SMG solvers should be updated to use the setbox hypre
methods for copying data.   This is a performance enhancement.

PR #756, opened 2026-08-05
Refactor Hypre Interface

mfahdaz:
## Summary

Refactor `pf_hypre.c` to replace element-wise Hypre Struct API calls with box-based operations.

As noted in issues #106, #251, #745, vector transfers and matrix assembly issued one Hypre API call per grid cell (`SetValues`,`GetValues`) i.e. from `BoxLoopI1` , creating significant function-call and GPU kernel-launch and synchronization overhead.

### Changes

* Replaced cell-wise vector transfers with one box-based transfer per subgrid.
* Replaced cell-wise matrix assembly with box-based insertion per stencil entry/subgrid.
* Followed the conventions used in `pfmgoctree.c` e.g. `value_box` , `set_box`.
* No numerical algorithm changes.

## Testing and Results

Performance testing of `PFMG` on CPUs and GPUs using ClayL testcase show:

* Significant reduction in total runtime and solver runtime.
* Hypre vector-copy overhead reduced significanlty.
* Performance comparable to `mgsemi` on GPUs.
* Better performance than `mgsemi` in CPU tests.

### DIFF
--- a/pfsimulator/parflow_lib/pf_hypre.c
+++ b/pfsimulator/parflow_lib/pf_hypre.c
@@ -1,6 +1,6 @@
 /*BHEADER**********************************************************************
 *
-*  Copyright (c) 1995-2024, Lawrence Livermore National Security,
+*  Copyright (c) 1995-2026, Lawrence Livermore National Security,
 *  LLC. Produced at the Lawrence Livermore National Laboratory. Written
 *  by the Parflow Team (see the CONTRIBUTORS file)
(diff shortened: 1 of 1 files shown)


smithsg84:
I'm confused by this PR.   The difference between PFMG and PFMG_Octree was using the octree loop and the box interface to hypre.  PFMG was using the element API.   The hypre box optimizations should be done in the Octree version not PFMG.   Is this simply copying the existing optimization over?

Is there a performance difference with PFMG_Octree?

PR #757, opened 2026-08-06
Fix integer truncation in CUDA/Kokkos memory allocation macros

mfahdaz:
## Summary

This PR fixes a long standing bug i.e. integer truncation in the CUDA/Kokkos allocation macros that can cause large allocation requests to be under-allocated before being passed to the underlying allocator.

The allocation macros narrow the requested element count to `unsigned int` before computing the allocation size. On systems where `size_t` is 64-bit, allocations with element counts exceeding the 32-bit range are truncated, resulting in a smaller buffer than requested.

The resulting out-of-bounds writes can trigger illegal memory access failures at unrelated locations. The observed crash location (e.g., `BoxloopI1()` in `matrix.c` line [769](https://github.com/parflow/parflow/blob/89360af061b4da1dd622632981379a372f4673fa/pfsimulator/parflow_lib/matrix.c#L769)) is therefore a downstream symptom of an earlier under-allocation.

This bug affects both native CUDA and Kokkos backends because they share the affected allocation patterns. The issue is only triggered when an individual allocation exceeds the 32-bit limit. Runs with smaller per-rank or per-device allocations are unaffected, which explains why the problem is not observed consistently across all GPU platforms.

The same truncation pattern is present in all amps communication backends `amps.h` files and also the CUDA and Kokkos allocation macros in:

- `pf_cudamalloc.h`
- `pf_kokkosmalloc.h`

## Changes/ Fix

Replaced the narrowing `unsigned int` casts with `size_t` byte-count calculations in all affected allocation macros.

No changes are required in the allocator implementations because the underlying allocation APIs (`cudaMallocManaged`, RMM, Umpire, and Kokkos allocation routines) already use `size_t` for allocation sizes.

## Validation

The fix was validated using ClayL with different sizes upto  `1584 × 1584 × 240` problem running on 4 GH200 GPUs.

- Before: reproducible illegal memory access failure.
- After: run completes successfully without errors.


smithsg84 (approved):
Thanks for bugfix

PR #758, opened 2026-09-01
Fix: populate PatchIndexOfDomainTop unconditionally in SetupRichards

rlsandovalp:
ComputePatchTop() is the only function that populates ProblemDataPatchIndexOfDomainTop(), a per-column array recording which named top-of-domain BC patch each surface cell belongs to. It was only called when public_xtra->print_top || public_xtra->write_silo_top was set, i.e. it looked like an output-only diagnostic.

When ComputePatchTop() is skipped, the vector is left at its allocation-time zero-fill instead of the real per-cell patch indices (which ComputePatchTop() explicitly starts from -1 before assigning), so every surface cell is silently treated as patch index 0 in the seepage-patch and mixed-surface-BC-mixing logic.

Verified by comparing KINSOL residual logs across three independent job submissions with and without this fix: the pre-fix residual matched a known-buggy baseline to ~13 significant digits at the first nonlinear solve, and the fixed build matched a known-correct baseline (same physical setup, run with Solver.PrintTop enabled so the bug was masked) to full precision from nni=0 onward.

The ComputePatchTop()  guards masked additional errors in the internal logic for domains with NZ=1, the index top can be 0 not >0 as was previously checked for.   This resulted in incorrect top patch index calculations.

A NZ=1 test was modified to write and check patch top and patch zindex for correctness.


smithsg84 (approved):
Thanks for tracking this down.

smithsg84 (changes requested):
This exposes some other issues in ParFlow and needs work.

smithsg84:
I have fixed the primary error in the tests (1) introduced by the initial guard removal for ComputeTopPatch().  The other comments identified should be checked.

Why PR #758 fails on some macOS CI tests

The fix exposes an out-of-bounds read that glibc hides and the macOS allocator does not.

ComputePatchTop() calls InitVectorAll(patch_top, -1) and then only assigns a patch index where index_top_data[index2d] > 0.  That test is > 0, not >= 0.  ComputeTopAndBottom() stores top_data = k for the top active cell, so any domain whose top is at k = 0 never gets a patch index and patch_top stays -1 everywhere.  That is every single-layer test: overland_tiltedV_KWE, overland_slopingslab_KWE, overland_FlatICP, overland_flux, overland_tiltedV_MGSemiTest, and reservoir_mpi_test all set ComputationalGrid.NZ = 1.  Ghost cells at the exterior boundary are also -1, since VectorUpdateAll has nothing to fill them from.

Before #758 none of this mattered.  The vector was calloc'd, every cell read patch 0, and IsSeepagePatch() indexed seepage_patches[0], a valid element.  After #758 those cells carry -1.  IsSeepagePatch() is just "return seepage->seepage_patches[patch_id]", so it reads seepage_patches[-1].  On glibc that address holds the upper four bytes of the 8-byte malloc chunk size, which is zero for any small allocation, so Linux reliably reads "not a seepage patch" and matches the old baseline.  On macOS the allocator keeps no inline header, so [-1] is the tail of whatever was allocated before it.  When that is nonzero, the seepage branch switches on for the whole surface, q_overlnd changes, and the top-of-domain pressures differ.  Which tests fail depends on heap layout, so only some fail.

There is a second consumer of the -1 values in overlandflow_eval_Kin.c.  It reads p0x = patch_dat[ipat - 1] and p0y = patch_dat[ipat - sy_v].  At exterior boundaries those are now ghost -1 while p1 is real, so the "p1 != p0x" internal-patch-edge branch fires at every domain edge and indexes pp through k1x = top_dat[itop + 1], which is also -1 in the ghost.  Before the fix, patch was 0 everywhere including ghosts, so that branch never fired.

Suggested changes to the PR

1. compute_patch_top.c: change the test to "if (index_top_data[index2d] >= 0)".  This fixes the NZ = 1 cases.

2. seepage.h: bounds-check IsSeepagePatch() so patch_id < 0 or patch_id >= num_patches returns 0.  This is the line that hides the platform difference.

3. overlandflow_eval_Kin.c: treat a negative neighbor patch as "no edge", i.e. require p0x >= 0 && p1 >= 0 (and the same for p0y) before entering the edge branch.

Fastest confirmation: build with -fsanitize=address and run one failing macOS test.  ASan should flag the seepage_patches[-1] read on the first nonlinear iteration.  If the failing list is the NZ = 1 KWE tests, the > 0 check alone explains it.

reedmaxwell:
Claude and I built this PR head (b1a9138e) with `-fsanitize=address` on macOS (arm64, Apple clang 16) and ran a real problem with it: a CONUS2.1 subset of the East River headwaters (32 x 41 x 10, masked solid file, OverlandKinematic top BC, coupled CLM, analytic Jacobian).  Three runs, zero ASan reports:

- the full coupled WY2017 year, 8760 hourly steps
- a P-ET spin-up to steady state under a SeepageFace top BC (8110 steps), which drives the seepage-patch machinery directly
- a 48 h startup probe

So the `>= 0` fix plus the merged #752 guard hold up clean under ASan on a masked watershed.  One caveat: ASan cannot see the remaining edge reads @smithsg84 flagged.  With k = -1 the computed index lands inside the allocated ghost layer, so these are in-bounds reads of wrong values.  That is also why glibc hides them.

Following Steve's items 2 and 3, I had Claude go through `overlandflow_eval_Kin.c` and the problem is wider than the interaction with this PR (this is somehting I've seen cause issues on Linux/Ubuntu builds previously):

1. #603 guarded the neighbor pressure reads in CALCFCN only.  CALCDER still reads `pp[ipp1]` / `pp[ippsy]` unguarded, with `k1x` / `k1y` taken raw from ghost `top_dat`.  The Jacobian then disagrees with the residual at east and north domain edges whenever the ghost pressure is positive.  This predates #758.
2. The internal-patch-edge branches from #585 read `pp[ipm1]` / `pp[ipmsy]` unguarded, in both CALCFCN and CALCDER.  With this PR populating the patch vector unconditionally, they fire at every exterior and mask edge (ghost patch = -1), not only at true internal edges.  The wrong flux enters the residual through `kw_v[io] = qx_v[io - 1]`, and its derivative enters JC through `ke_der` / `kw_der`.  Newton stays consistent with the wrong residual and converges with no solver symptom; only the answer moves.  I believe that is the silent baseline drift in the macOS CI runs.
3. Those branches index the i-1 / j-1 upwind cell with the wrong top value: they use `k1x` / `k1y` (the tops of i+1 / j+1) where they need `k0x` / `k0y`.
4. This fix didn't actually cause any errors (see below), meaning the edge values are assigned default in such a way that they don't change the Jacobian or preconditioner for my East River test so the results with and without this change were identical (pressure values and solver / kinsol logs), but they did cause intermittent fails on an Ubuntu build and they could cause issues for some types of domains (I think, not confirmed).

The fix is on [`bugfix_calcder`](https://github.com/parflow/parflow/tree/bugfix_calcder) (693641f3, one commit on master), tested A/B on the East River year: guard the CALCDER reads the way #603 guarded CALCFCN, but on the top value (`k1x >= 0`) rather than the computed index, which can land in a ghost layer and pass a sign check; enter the patch-edge branches only when `p1 >= 0 && p0 >= 0` (Steve's item 3); and use `k0x` / `k0y`.  Diff below.  The guarded binary reproduces the unguarded year bit for bit: identical end-state md5 and byte-identical KINSOL logs across all 8760 steps.  That is expected as I mentioned above in (4).  The East River mask edges are watershed divides and never pond, so every wrong-value read clamps to zero through `pfmax(., 0)`.  The configurations that can tell the two binaries apart are the NZ = 1 KWE tests and any domain that ponds at an edge, which is the macOS CI failure set.

I am happy to open `bugfix_calcder` as a separate PR, or it can fold in here.  It does not include the `IsSeepagePatch()` bounds check (Steve's item 2); that belongs with this PR.

<details><summary>the diff (same as https://github.com/parflow/parflow/compare/master...bugfix_calcder)</summary>

```diff
diff --git a/pfsimulator/parflow_lib/overlandflow_eval_Kin.c b/pfsimulator/parflow_lib/overlandflow_eval_Kin.c
index 454eb9ca..0d1ca95f 100644
--- a/pfsimulator/parflow_lib/overlandflow_eval_Kin.c
+++ b/pfsimulator/parflow_lib/overlandflow_eval_Kin.c
@@ -162,9 +162,9 @@ void    OverlandFlowEvalKin(
         PP_ipp1 = 0.0;
         PP_ippsy = 0.0;
         PP_ip = pp[ip];
-        if (ipp1 >= 0)
+        if (k1x >= 0)
           PP_ipp1 = pp[ipp1];
-        if (ippsy >= 0)
+        if (k1y >= 0)
           PP_ippsy = pp[ippsy];
         Press_x = RPMean(-Sf_x, 0.0,
                          pfmax((PP_ip), 0.0),
@@ -179,14 +179,14 @@ void    OverlandFlowEvalKin(
                              * mann_dat[io])) * RPowerR(Press_y, (5.0 / 3.0));
       }
       // fix for internal patch edges in x direction
-      if (p1 != p0x)
+      if (p1 >= 0 && p0x >= 0 && p1 != p0x)
       {
         if (k1 >= 0)
         {
           ip = SubvectorEltIndex(p_sub, i, j, k1);
           Sf_x = sx_dat[io - 1];
           Sf_y = sy_dat[io - 1];
-          ipm1 = (int)SubvectorEltIndex(p_sub, i - 1, j, k1x);
+          ipm1 = (int)SubvectorEltIndex(p_sub, i - 1, j, k0x);
 
           Sf_mag = RPowerR(Sf_x * Sf_x + Sf_y * Sf_y, 0.5);
           if (Sf_mag < ov_epsilon)
@@ -202,14 +202,14 @@ void    OverlandFlowEvalKin(
       }
 
       // fix for internal patch edges in y direction
-      if (p1 != p0y)
+      if (p1 >= 0 && p0y >= 0 && p1 != p0y)
       {
         if (k1 >= 0)
         {
           ip = SubvectorEltIndex(p_sub, i, j, k1);
           Sf_x = sx_dat[io - sy_v];
           Sf_y = sy_dat[io - sy_v];
-          ipmsy = (int)SubvectorEltIndex(p_sub, i, j - 1, k1y);
+          ipmsy = (int)SubvectorEltIndex(p_sub, i, j - 1, k0y);
 
           Sf_mag = RPowerR(Sf_x * Sf_x + Sf_y * Sf_y, 0.5);
           if (Sf_mag < ov_epsilon)
@@ -299,7 +299,8 @@ void    OverlandFlowEvalKin(
                                          int k1, k0x, k0y, k1x, k1y;
                                          int p1, p0x, p0y;
                                          double Sf_x, Sf_y, Sf_mag;
-                                         double Press_x, Press_y, qx_temp, qy_temp; ),
+                                         double Press_x, Press_y, qx_temp, qy_temp;
+                                         double PP_ipp1, PP_ippsy; ),
                                   CellSetup(DoNothing),
                                   FACE(LeftFace, DoNothing), FACE(RightFace, DoNothing),
                                   FACE(DownFace, DoNothing), FACE(UpFace, DoNothing),
@@ -333,12 +334,19 @@ void    OverlandFlowEvalKin(
         if (Sf_mag < ov_epsilon)
           Sf_mag = ov_epsilon;
 
+        PP_ipp1 = 0.0;
+        PP_ippsy = 0.0;
+        if (k1x >= 0)
+          PP_ipp1 = pp[ipp1];
+        if (k1y >= 0)
+          PP_ippsy = pp[ippsy];
+
         Press_x = RPMean(-Sf_x, 0.0,
                          pfmax((pp[ip]), 0.0),
-                         pfmax((pp[ipp1]), 0.0));
+                         pfmax((PP_ipp1), 0.0));
         Press_y = RPMean(-Sf_y, 0.0,
                          pfmax((pp[ip]), 0.0),
-                         pfmax((pp[ippsy]), 0.0));
+                         pfmax((PP_ippsy), 0.0));
 
         qx_temp = -(5.0 / 3.0) * (Sf_x / (RPowerR(fabs(Sf_mag), 0.5) * mann_dat[io])) * RPowerR(Press_x, (2.0 / 3.0));
         qy_temp = -(5.0 / 3.0) * (Sf_y / (RPowerR(fabs(Sf_mag), 0.5) * mann_dat[io])) * RPowerR(Press_y, (2.0 / 3.0));
@@ -350,14 +358,14 @@ void    OverlandFlowEvalKin(
       }
 
       // fix for internal patch edges in x direction
-      if (p1 != p0x)
+      if (p1 >= 0 && p0x >= 0 && p1 != p0x)
       {
         if (k1 >= 0)
         {
           ip = SubvectorEltIndex(p_sub, i, j, k1);
           Sf_x = sx_dat[io - 1];
           Sf_y = sy_dat[io - 1];
-          ipm1 = (int)SubvectorEltIndex(p_sub, i - 1, j, k1x);
+          ipm1 = (int)SubvectorEltIndex(p_sub, i - 1, j, k0x);
 
           Sf_mag = RPowerR(Sf_x * Sf_x + Sf_y * Sf_y, 0.5);
           if (Sf_mag < ov_epsilon)
@@ -374,14 +382,14 @@ void    OverlandFlowEvalKin(
       }
 
       // fix for internal patch edges in y direction
-      if (p1 != p0y)
+      if (p1 >= 0 && p0y >= 0 && p1 != p0y)
       {
         if (k1 >= 0)
         {
           ip = SubvectorEltIndex(p_sub, i, j, k1);
           Sf_x = sx_dat[io - sy_v];
           Sf_y = sy_dat[io - sy_v];
-          ipmsy = (int)SubvectorEltIndex(p_sub, i, j - 1, k1y);
+          ipmsy = (int)SubvectorEltIndex(p_sub, i, j - 1, k0y);
 
           Sf_mag = RPowerR(Sf_x * Sf_x + Sf_y * Sf_y, 0.5);
           if (Sf_mag < ov_epsilon)
```

</details>

